PR #243, opened 2026-04-15
fix(file-browser): default to 75% upper split on launch (#237)

### LINKED ISSUE #237: File explorer: default to 75% upper split on open
## Request

When the file explorer opens, it should default to taking 75% of the upper split, not whatever it's currently doing (50/50 or full pane depending on how it was launched).

## Reproduce

Open the file explorer from any launch path (palette, Finder Service, CLI). Observe the default split ratio.

## Expected

File explorer lands as a 75% upper split by default.

## Scope hint

This is likely either:
- A field in \`manifest.toml\` for the file-browser app under \`[app.launch]\` (something like \`companion_size = 0.25\`).
- A hard-coded default in \`src/pane_ops.rs\` for the file-browser launch path.

Either way, this is a config/default change, not a protocol change. The user suspects it may be a config file toggle — check there first.

## Source

Backlog note \`~/.plexi/backlog/note-2026-04-14-234240.md\`

ianjamesburke:
## Summary

Pin the built-in file browser to a 75% upper split (app = 75%, companion terminal = 25%) at the `open_file_browser` launch site so it lands deterministically regardless of future drift in `DEFAULT_COMPANION_FRACTION` or unrelated manifest defaults.

## Scenario

**C (narrow variant).** The file browser is a built-in Rust app (`src/file_browser/mod.rs`) with no manifest — Scenario A/B (manifest field) don't apply. The only launch site is `PlexiApp::open_file_browser` in `src/pane_ops.rs`, so the fix is scoped to that single function.

## Change

`src/pane_ops.rs` — 9 lines added after `open_app_on_focused` in `open_file_browser`. After the app is embedded, we look up the focused pane and, if its `SurfaceMode` is `AppWithCompanion`, overwrite `companion_fraction` to `0.25`. No other app launch path is touched.

```rust
// Pin file browser to a 75% upper split (companion = 25%). See #237.
if let Some((_id, pane)) = self.contexts[self.active_context].focused_pane_mut() {
    if let crate::app_trait::SurfaceMode::AppWithCompanion { ref mut companion_fraction } =
        pane.surface_mode
    {
        *companion_fraction = 0.25;
    }
}
```

## Test plan

- [ ] Launch the file browser from a blank Plexi window — verify it takes the top 75% with the terminal at the bottom 25%.
- [ ] Launch the file browser from a pane that already has a terminal with history — verify the terminal is preserved and now occupies the bottom 25%.
- [ ] Open wikipedia / git-log — verify their split ratios are unchanged from current behavior.
- [ ] `cargo check` passes (confirmed locally — 42 unrelated warnings, no errors).

Closes #237

### DIFF
--- a/src/pane_ops.rs
+++ b/src/pane_ops.rs
@@ -781,6 +781,15 @@ impl PlexiApp {
         // Built-in file browser gets full permissions.
         let perms = crate::app_permissions::AppPermissions::builtin();
         self.open_app_on_focused(app, perms, cwd);
+
+        // Pin file browser to a 75% upper split (companion = 25%). See #237.
+        if let Some((_id, pane)) = self.contexts[self.active_context].focused_pane_mut() {
+            if let crate::app_trait::SurfaceMode::AppWithCompanion { ref mut companion_fraction } =
+                pane.surface_mode
+            {
+                *companion_fraction = 0.25;
+            }
+        }
     }
 
     /// Open an app on the focused pane WITHOUT creating a linked terminal split.


ianjamesburke:
Closing — superseded by the launch-mode manifest fix in alpha commit f8da18e (src/pane_ops.rs). The new approach is more general: any app declares [app.launch] mode = "fullscreen" / companion = "none" in its manifest and the host honors it. The file browser can opt into the 75% upper split by setting companion = "terminal" with companion_size = 0.25 in its manifest — no special-case code needed. If you want to wire that manifest update on the file browser specifically, file a fresh issue against alpha.

PR #290, opened 2026-04-22
fix: Cmd+W on last pane resets context instead of destroying it

ianjamesburke:
## Summary
- Previously, Cmd+W with a single pane in a multi-context workspace would call `delete_context()`, removing the context from the sidebar
- Now always calls `reset_active_context()` when the last pane is closed, spawning a fresh blank terminal in the same context
- Context deletion via Cmd+W is no longer possible — only the sidebar X button removes a context

## Test plan
- [ ] Open Plexi with multiple contexts in the sidebar
- [ ] Focus a context that has only one pane, press Cmd+W
- [ ] Context should remain in sidebar with a fresh blank terminal — not disappear
- [ ] Open multiple panes in a context, Cmd+W them down to the last one — same reset behavior
- [ ] Sidebar X button still deletes a context as expected


ianjamesburke:
Superseded by #289 — execute_close_pane() in the config/close-safety PR already implements the blank-pane behavior (no delete_context path).

PR #419, opened 2026-04-29
fix(pyright): zero errors across examples and SDK

ianjamesburke:
## Summary

- `iq_query_test.py`: `ctx.rect` was being used as a subscriptable dict — it's a draw method. Replaced with `ctx.x`, `ctx.y`, `ctx.w`.
- `app_store.py`: `ctx.width` doesn't exist on `RenderContext`; replaced with `ctx.w`. Also removed the `hasattr` guard — `ctx.w` is always present.
- `audio_bridge_test.py`, `canvas_terminal_bindings_test.py`: `ctx.schedule_render()` doesn't exist; `schedule_render` lives on `Emitter`. Fixed to `self.emit.schedule_render()`.
- `iq_query_denied_test.py`: `_state` was typed as `object`, making subscripting and unpacking invalid. Replaced with the correct union `str | tuple[str, str] | None`.
- `mind_map.py`: `node.parent` is `int | None`; added `assert node.parent is not None` before use (the enclosing `if self._selected != self._root` guard makes this invariant true, the assert narrows the type).
- `sdk/python/plexi_sdk/__init__.py`: `measure_text()` was missing a return path when stdin closes before a matching `TextMeasured` response arrives; added `raise RuntimeError(...)` so Pyright sees all paths return.

**Breaks if:** `pyright` (run from the repo root) reports errors on `examples/` or `sdk/python/`.

## Test plan
- [ ] `pyright` from repo root: 0 errors, 0 warnings

### DIFF
--- a/examples/app-store/app_store.py
+++ b/examples/app-store/app_store.py
@@ -812,7 +812,7 @@ class AppStoreApp(App):
             "app-store-url",
             x=20.0,
             y=input_y,
-            w=ctx.width - 40.0 if hasattr(ctx, "width") else 600.0,
+            w=ctx.w - 40.0,
             placeholder="github:owner/repo[@ref]  or  git+https://…  (Enter to install)",
         )
         if submitted is not None:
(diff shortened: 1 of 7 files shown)


ianjamesburke:
Closing in favour of #437, which cherry-picks the valid fixes onto current alpha. The original PR conflicts due to the IQ→AI rename (#433) and async SDK rewrite. Two fixes were already resolved upstream: ctx.rect fix landed in #433, and measure_text return path is inapplicable to the current async implementation.

Issue #429, opened 2026-04-30
Refactor: Rename IQ → AI, modular backend (OpenRouter + Ollama), delete legacy Claude CLI path


ianjamesburke:
## Current State (audited 2026-05-01)

Most of this work is already done. The only remaining task is deleting the orphaned `src/plexi_iq/` directory.

| Phase | Status | Notes |
|---|---|---|
| Phase 1 — Rename IQ → AI | ✅ Done | `plexi_ai/` module live, `AiConfig` in config.rs, `mod plexi_iq` removed from `main.rs` |
| Phase 2 — Ollama backend + modular config | ✅ Done | `plexi_ai/backend/ollama.rs` exists and is wired in `broker.rs` via `dispatch_ollama()` |
| Phase 3 — HTTP timeouts | ✅ Done | 10s connect + 30s overall in `openrouter.rs:111-112`; Ollama uses 60s |
| Phase 4 — Delete legacy paths | ✅ Done | `LlmRequest`/`LlmResponse` gone from protocol and routing; `emit.llm()` not found in SDK |
| Phase 5 — Ledger rename | ✅ Done | `ledger_path()` in `plexi_ai/ledger.rs` writes to `ai-ledger.jsonl` with migration from `agent-ledger.jsonl` |
| **Delete `src/plexi_iq/`** | ❌ **TODO** | Directory still on disk, not compiled (no `mod plexi_iq` in `main.rs`), but ~40KB of orphaned dead code |

**Only remaining action:** `rm -rf src/plexi_iq/` and verify `cargo build` passes.

---

## What

Plexi's LLM infrastructure is fragmented across **four overlapping paths**, uses stale internal jargon ("IQ") everywhere, is hardcoded to a single OpenRouter backend, and has zero HTTP timeout safety. This issue scopes a unified refactor that:

1. **Renames "IQ" → "AI"** across the entire codebase — protocol, config, modules, logs, file paths.
2. **Deletes the legacy `agent_pane.rs` in-process Claude CLI backend** and `agent_turn.rs` entirely.
3. **Introduces a pluggable `AiBackend` trait** with OpenRouter (cloud) and Ollama (local) as first-class backends.
4. **Unifies all LLM paths into one** — everything routes through the brokered `ai.query` capability.
5. **Adds 30-second HTTP timeouts** to all LLM-related `ureq` calls.
6. **Breaks backward compatibility freely** — no active users, no migration path needed.

### Current fragmentation (four paths → one)

[...]

PR #430, opened 2026-04-30
feat(install): detect and append missing config sections on install

### LINKED ISSUE #425: just install-alpha: detect and migrate missing config sections
## Problem

When new required config sections are added (e.g. `[iq]` in v3.3), existing users' `~/.plexi-alpha/config.toml` files are silently missing them. The install process overwrites the binary but never touches the config. The result is silent runtime failures that are hard to diagnose.

## Proposed fix

Add a `migrate-config` step to `just install-alpha` (and `install-beta`, `install`) that:

1. Reads the canonical `CONFIG_TEMPLATE` string from the compiled binary (or a sidecar file)
2. Parses the user's existing config
3. For each top-level section in the template that is absent from the user's config, appends it (with all keys commented-out and a note: `# Added by installer — uncomment and fill in`)
4. For required sections with no safe default (like `[iq]`), appends with actual default values and a comment directing the user to set their API key
5. Prints a summary of what was added: `config: added [iq] section (see ~/.plexi-alpha/config.toml)`

## Non-goals

- Never modify values the user has set
- Never remove sections the user has added
- Not a full config schema validator — just an additive patch for missing sections

## Implementation

Could be a small Rust binary (`plexi-alpha --migrate-config`) called from the Justfile, or a shell script that uses `grep` to check for section headers and `cat >>` to append missing blocks. The shell script approach is simpler and has no compile dependency.

## Related

Discovered during v3.3 smoke testing: `[iq]` section missing from existing alpha config caused IQ queries to hang silently with no user-visible error. See also #424 (error visibility).

## Merge Checklist

- [ ] On a machine with an existing config missing a new section: run `just install-alpha` — the missing section is added, existing values are preserved
- [ ] On a fresh install with no config: the generated config contains all required sections
- [ ] Running install twice does not duplicate any config section

ianjamesburke:
Closes #425

## Summary
- `scripts/migrate-config.sh` accepts a config path argument and appends any missing sections with defaults and comments
- Checks for both `[ai]` and legacy `[iq]` — users who already upgraded via #429 don't get a duplicate section
- `just install-alpha`, `install-beta`, `install-v3`, and `install` all call the script for their respective config paths
- Standalone `migrate-config-alpha`, `migrate-config-beta`, `migrate-config-v3` recipes for manual invocation
- Script is idempotent — running twice prints "up to date — no sections added"
- Silently skips if no config.toml exists yet (first install)

## Test plan
- [ ] Remove `[iq]`/`[ai]` section from `~/.plexi-alpha/config.toml`, run `just install-alpha`, verify `[ai]` section is appended with correct defaults
- [ ] Run `just install-alpha` again — "up to date — no sections added" printed, no duplicate section
- [ ] With existing `[iq]` section present — "up to date" printed, no `[ai]` section added
- [ ] `cargo build` passes clean

🤖 Generated with Claude Code

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,6 +12,17 @@ install:
     cp target/release/bundle/osx/Plexi.app/Contents/MacOS/plexi /usr/local/bin/plexi
     rm -rf /Applications/Plexi.app
     cp -r target/release/bundle/osx/Plexi.app /Applications/Plexi.app
+    @bash scripts/migrate-config.sh "$HOME/.plexi/config.toml"
+
+# Append any missing config sections to a Plexi config.toml. Idempotent.
+migrate-config-alpha:
+    @bash scripts/migrate-config.sh "$HOME/.plexi-alpha/config.toml"
+
+migrate-config-beta:
+    @bash scripts/migrate-config.sh "$HOME/.plexi-beta/config.toml"
+
+migrate-config-v3:
+    @bash scripts/migrate-config.sh "$HOME/.plexi-v3/config.toml"
 
 # ── Versions — full lifecycle for parallel .app installs + worktrees ─────────
 #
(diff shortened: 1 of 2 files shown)


ianjamesburke:
Closing — not needed right now.

PR #573, opened 2026-05-03
feat(zoom): zoomed pane at 99% opacity

### LINKED ISSUE #572: feat: zoomed pane overlay at 99% opacity
When a pane is zoomed to full-screen, render the overlay frame at 88% opacity so the dimmed background is faintly visible through it.

**Implementation:** In `app/mod.rs` zoom overlay section, call `child_ui.set_opacity(0.88)` before the `egui::Frame::show()` call. This is the same pattern used by the minimap (`ui.set_opacity(0.75)` in `src/minimap.rs`) — it multiplies all painter calls including the terminal's internal cell background renders.

**Do NOT** change the `Frame::fill` color alpha. The terminal renderer inside the Frame paints its own fully-opaque cell backgrounds on top of the fill, overriding any alpha set there. Only `set_opacity` propagates through the entire paint chain.

**Breaks if:** Zooming any pane shows the overlay frame as fully transparent or fully opaque with no visible bleed-through from the scrim.

## Prior Attempts

**Attempt 1:** Changed `egui::Frame` fill from fully-opaque `terminal_bg` to an alpha-adjusted color (252/255 alpha).
**Why it failed:** Frame fill alpha is overridden by the terminal renderer's own fully-opaque cell background paints on top.
**What to try next:** Use `set_opacity` on the child_ui instead.

**Attempt 2:** `child_ui.set_opacity(zoom_opacity)` with default 0.98, plus a configurable `beta.zoom_opacity` key in `config.toml`.
**Why it failed:** At 98–99% opacity the difference from 100% is visually imperceptible, especially with a dark terminal bg against a dark scrim.
**What to try next:** Use a more meaningful opacity drop — 0.85–0.92 range — to make bleed-through actually visible. 0.95 (tried in PR #618) also had no visible difference.

ianjamesburke:
## Summary
- Zoomed (full-screen) pane frame fill is now `252/255` alpha instead of fully opaque
- Lets 1% of the dimmed background scrim bleed through — barely visible, subtle depth effect
- No new dependencies, 2-line change in `app/mod.rs`

Closes #572

## Test plan
- [ ] Open any pane, zoom it (`⌘Z` or equivalent)
- [ ] Verify the background panes are faintly visible through the zoomed overlay (not a solid wall)
- [ ] Verify terminals, apps, and agent panes all work normally while zoomed

**Breaks if:** The zoomed overlay is fully opaque with zero bleed-through, or fully transparent.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2068,9 +2068,9 @@ impl eframe::App for PlexiApp {
                             !i.raw.dropped_files.is_empty()
                                 && child_ui.rect_contains_pointer(inner_rect)
                         });
-                        let bg = self.colors.terminal_bg;
+                        child_ui.set_opacity(self.features.zoom_opacity);
                         egui::Frame::new()
-                            .fill(Color32::from_rgba_unmultiplied(bg.r(), bg.g(), bg.b(), 252))
+                            .fill(self.colors.terminal_bg)
                             .inner_margin(egui::Margin::same(8))
                             .show(&mut child_ui, |ui| {
                                 if let Some(pane) = ctx.panes.get_mut(&pane_id) {
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2068,8 +2068,9 @@ impl eframe::App for PlexiApp {
                             !i.raw.dropped_files.is_empty()
                                 && child_ui.rect_contains_pointer(inner_rect)
                         });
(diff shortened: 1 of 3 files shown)


ianjamesburke:
Closing — approach didn't produce the intended visual effect. See issue #572 for full failure notes.

PR #587, opened 2026-05-03
fix(clipboard): drop \n for client-wrapped terminal rows

### LINKED ISSUE #301: v3.1: Terminal clipboard has embedded \n for client-wrapped lines
## Symptom

Selecting a multi-line run of text in a Plexi terminal pane (e.g. a paragraph of agent output) and copying yields a string with embedded `\n` at every display-row boundary, even when the source text was logically a single line that the emitter hard-wrapped to fit the terminal width.

Concrete repro: in Claude Code (or any tool that does its own column-width wrapping via `COLUMNS`), print a long paragraph. Select and copy → clipboard contains the paragraph broken at every display row.

## Root cause

Many CLI tools detect terminal width via `tcgetwinsz` / `$COLUMNS` and manually insert `\n` at the wrap point before the text reaches the PTY. Our terminal backend (alacritty-based) correctly treats those as hard line breaks — `CellFlags::WRAPLINE` is never set because the row wasn't soft-wrapped by the terminal.

`deps/egui_term/src/backend/mod.rs::selectable_content` (~line 282) already checks WRAPLINE and only inserts `\n` when the previous row was not soft-wrapped. The logic is correct; it just has no way to distinguish a legitimate hard line break from a client-inserted wrap.

## Fix options

1. Heuristic re-glue at copy time: if two adjacent non-WRAPLINE rows look like a client wrap (row ends at right edge, next row starts with a non-space continuation), drop the `\n`. Fragile — risks gluing short adjacent sentences that happen to abut.
2. Platform convention: accept as-is. iTerm2, Terminal.app, kitty all produce the same artifact and users are accustomed to it.

Deferred to v3.1 so we can evaluate usage before shipping a heuristic.

## Related

A sibling bug (clickable URL underline only covers the first row of a wrapped URL) is being fixed separately on alpha because it's user-visible and higher impact. That fix applies only to link detection, not to the copy path.

ianjamesburke:
## What

When a CLI tool (e.g. Claude Code, git log) manually hard-wraps a paragraph to fit `COLUMNS`, those hard `\n` characters reach the PTY and the terminal stores them as genuine line breaks — no `WRAPLINE` flag is set. Selecting and copying across those rows previously produced clipboard text with `\n` at every display-row boundary.

## Fix

`selectable_content` in `deps/egui_term/src/backend/mod.rs` now checks the last cell of the completed row before inserting a `\n`. If the last column is non-space, the row filled to the terminal edge — a strong signal that a CLI tool wrapped the line. Skip the `\n`. Empty cells default to `' '`, so `last_cell.c == ' '` reliably identifies genuine line breaks shorter than terminal width.

## Trade-off

A genuine line that ends exactly at the terminal column boundary loses its `\n`. This edge case is rare in practice. All other terminal emulators (iTerm2, kitty, Terminal.app) produce the embedded-`\n` artifact instead — any improvement beats baseline.

Closes #301

### DIFF
--- a/deps/egui_term/src/backend/mod.rs
+++ b/deps/egui_term/src/backend/mod.rs
@@ -308,15 +308,18 @@ impl TerminalBackend {
                 if range.contains(indexed.point) {
                     if let Some(prev) = prev_line {
                         if prev != indexed.point.line {
-                            // Only insert a newline if the previous row was NOT
-                            // soft-wrapped (WRAPLINE flag on its last cell).
-                            let was_wrapped = content.grid[prev][last_col]
-                                .flags
-                                .contains(CellFlags::WRAPLINE);
+                            let last_cell = &content.grid[prev][last_col];
+                            let was_wrapped =
+                                last_cell.flags.contains(CellFlags::WRAPLINE);
                             // Trim trailing spaces from the completed line
                             let trimmed_len = result.trim_end_matches(' ').len();
                             result.truncate(trimmed_len);
-                            if !was_wrapped {
+                            // Skip \n for terminal soft-wraps (WRAPLINE) or for
+                            // rows that filled to the last column — a strong signal
+                            // that the CLI tool hard-wrapped the line itself (e.g.
+                            // Claude Code, git log). Empty last columns default to
+                            // ' ', so last_cell.c == ' ' means a genuine line break.
+                            if !was_wrapped && last_cell.c == ' ' {
                                 result.push('\n');
                             }
                         }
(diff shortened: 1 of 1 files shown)


ianjamesburke:
Reverting — research confirms no mainstream terminal emulator (iTerm2, kitty, WezTerm, Ghostty, Alacritty) solves this. Root cause: CLI tools (Claude Code, git log) emit hard \n before the PTY, making client-wrapped rows indistinguishable from genuine line breaks at the terminal level. No OSC protocol exists to signal soft vs hard wraps. Alacritty has WRAPLINE infrastructure but deliberately doesn't use it for clipboard. Heuristic (last-column non-space = client wrap) is too fragile to justify.

PR #591, opened 2026-05-03
fix(notifications): Cmd+Shift+A opens empty notification window when queue is empty

### LINKED ISSUE #590: fix(notifications): Cmd+Shift+A should open empty notification window when queue is empty
## Problem

Pressing Cmd+Shift+A when there are no pending notifications silently does nothing. The action is swallowed in the dispatch handler (`visible_notification_count() > 0` guard in `src/app/mod.rs`) and never opens the modal.

This feels broken — the user pressed a key and nothing happened, with no feedback.

## Expected

Cmd+Shift+A always opens the notification modal. When the queue is empty, the modal shows a simple "No notifications" empty state. Esc closes it.

## Implementation

Two changes:

1. **`src/app/mod.rs`** — Remove the `visible_notification_count() > 0` guard in the `ToggleNotificationModal` dispatch so it always opens the modal.

2. **`src/overlays.rs`** — In `draw_notification_modal`, when `pending_notifications.is_empty()`: instead of silently returning, render the scrim + a centered modal card showing a dim "No notifications" message. Handle Esc to close.

## Meta
depends_on: []

ianjamesburke:
Closes #590

## Summary
- Removes the `visible_notification_count() > 0` guard in `ToggleNotificationModal` dispatch — the modal now always opens on Cmd+Shift+A
- `draw_notification_modal` in `overlays.rs`: when queue is empty, renders scrim + centered card with dim "No notifications" label instead of silently returning. Esc closes it.

## Test plan
- [ ] Press Cmd+Shift+A with no notifications → modal opens with "No notifications"
- [ ] Press Esc → modal closes
- [ ] Press Cmd+Shift+A again → closes (toggle)
- [ ] Send a notification, press Cmd+Shift+A → normal notification modal opens as before

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1856,7 +1856,7 @@ impl eframe::App for PlexiApp {
                 Action::ToggleNotificationModal => {
                     if self.show_notification_modal {
                         self.show_notification_modal = false;
-                    } else if self.visible_notification_count() > 0 {
+                    } else {
                         self.show_notification_modal = true;
                         // Pick highest-priority when re-opening the modal and
                         // nothing is currently pinned. If something IS pinned
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2476,8 +2476,7 @@ impl PlexiApp {
     }
 
     pub(crate) fn sync_notification_modal_focus(&mut self) {
-        let should_own = self.show_notification_modal
-            && self.visible_notification_count() > 0;
+        let should_own = self.show_notification_modal;
         let has_layer = self
             .focus_stack
             .iter()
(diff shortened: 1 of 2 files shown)


ianjamesburke:
Closing — incomplete. Three guards found and fixed, but a fourth at src/app/mod.rs:2292 resets show_notification_modal = false every frame when the queue is empty. That one has a comment explaining it's intentional (avoids re-drawing the modal after the overlay path). Need to approach this differently — either pass show_notification_modal through to the render path as a distinct empty-state flag, or track "user explicitly opened" separately from "has notifications". Reopening fresh from the issue.

PR #763, opened 2026-05-06
fix(typed_pipes): detect drain thread broken pipe within one render frame

### LINKED ISSUE #759: fix(typed_pipes): drain thread exits silently on Broken Pipe — host detects stale session late
## What

When the app-side socket closes unexpectedly (e.g. the audio-recorder app switches devices mid-capture), the drain thread hits \`Broken pipe (os error 32)\` on \`write_frame\`, logs a warning, and exits silently. The host has no signal that the drain thread died — it only discovers the stale session the next time \`start_audio_capture\` is called with the same \`pipe_id\`. The gap is 3–21 seconds in practice.

Observed in the log:
\`\`\`
[WARN]  typed_pipes: drain write error: Broken pipe (os error 32)
[ERROR] app::audio-recorder pipe read_frame error: [Errno 9] Bad file descriptor
[INFO]  audio-recorder: capture stopped, 582 chunks
# ... 3–21 second gap ...
[WARN]  ProcessApp[audio-recorder]: AudioCapture pipe_id=audio-recorder-mic stale session — cleaning up and restarting
\`\`\`

## Reproduce

1. Open audio-recorder app
2. Start a capture, then switch audio devices while recording
3. Observe the log gap between the \`Broken pipe\` warn and the \`stale session\` warn

## Expected

The host should detect the broken drain within one render frame and immediately emit \`AudioCaptureError\` (and equivalent events for MIDI/video pipes), cleaning up the stale session proactively rather than waiting for the app to retry.

## Action Plan

- **\`src/typed_pipes.rs\`** — add \`error_flag: Arc<AtomicBool>\` to \`BinaryPipeEntry\`. Drain thread sets it before exiting with a write error. Add \`TypedPipeRegistry::drain_failed(pipe_id) -> bool\` to poll the flag.
- **\`src/process_app/routing.rs\`** — in \`poll_actions\` (or the per-frame tick path), iterate open audio/MIDI/video capture sessions and call \`drain_failed(pipe_id)\`. On true: close the pipe, remove the session from \`audio_capture_sessions\`, push \`PlexiEvent::AudioCaptureError\` so the app gets immediate notification.
- **\`src/typed_pipes.rs\`** — clear the flag in \`close()\` so it doesn't fire spuriously after a clean shutdown.

Alternative: pass a \`crossbeam_channel::Sender<(String, String)>\` (pipe_id, error) into the drain thread and let it push directly onto the host outbound event queue — more immediate, but requires routing the sender through \`open_binary\`.

## Done When

- After a Broken Pipe on the drain thread, \`AudioCaptureError\` (or equivalent) is emitted to the app within one render frame (~16ms), with no multi-second gap
- \`stale session\` warn is eliminated — the pipe is cleaned up before any retry attempt
- \`cargo test --lib\` passes

ianjamesburke:
Closes #759

## What

When the drain thread exits due to a write error (e.g. `Broken pipe os error 32`), the host had no signal — it only discovered the stale session on the next `start_audio_capture` retry, 3–21 seconds later.

## How

**`src/typed_pipes.rs`**
- Add `error_flag: Arc<AtomicBool>` to `BinaryPipeEntry`
- Drain thread sets it before breaking on a write error (alongside the existing `log::warn!`)
- New `TypedPipeRegistry::drain_failed(pipe_id) -> bool` polls the flag

**`src/process_app/mod.rs`**
- Per-frame check: after draining HTTP/file-picker async results, iterate all active `audio_capture_sessions` and call `drain_failed()` on each pipe
- On failure: remove session, close pipe, clear peak meter, push `PlexiEvent::AudioCaptureError` immediately
- Eliminates the `stale session` warn — pipe is cleaned up before any retry

## Tests

- `drain_failed_set_after_broken_pipe` — connects/drops a client socket, pushes a frame, asserts `drain_failed()` returns true within 300ms
- `drain_failed_false_for_healthy_pipe` — asserts flag is false before any error and for unknown pipe ids

## Breaks if

`AudioCaptureError` is not emitted within ~one render frame (~16ms) after a `Broken pipe` on the drain thread, or the `stale session — cleaning up and restarting` warn still appears in the log after a device switch.

### DIFF
--- a/src/process_app/mod.rs
+++ b/src/process_app/mod.rs
@@ -1291,6 +1291,41 @@ impl App for ProcessApp {
             self.outbound_events.push_back(event);
         }
 
+        // Per-frame: detect audio capture pipes whose drain thread exited due
+        // to a write error (e.g. Broken pipe when the app-side socket closed
+        // unexpectedly). Emit AudioCaptureError immediately so the app is
+        // notified within one render frame rather than waiting for the next
+        // start_audio_capture retry to discover the stale session.
+        let failed_audio_pipes: Vec<String> = self
+            .audio_capture_sessions
+            .keys()
+            .filter(|pipe_id| {
+                self.pipe_registry
+                    .lock()
+                    .map(|r| r.drain_failed(pipe_id))
+                    .unwrap_or(false)
+            })
+            .cloned()
+            .collect();
+        for pipe_id in failed_audio_pipes {
+            log::warn!(
+                "ProcessApp[{}]: AudioCapture pipe_id={pipe_id} drain failed — cleaning up",
+                self.type_id
+            );
+            self.audio_capture_sessions.remove(&pipe_id);
+            self.pipe_registry
+                .lock()
+                .expect("pipe_registry poisoned")
+                .close(&pipe_id);
+            if let Ok(mut m) = self.audio_peak_meters.lock() {
+                m.remove(&pipe_id);
+            }
+            self.outbound_events.push_back(PlexiEvent::AudioCaptureError {
+                pipe_id,
+                error: "pipe drain failed (broken pipe)".to_owned(),
+            });
+        }
+
         let new_cmds = self.drain_draw_commands();
 
         for cmd in new_cmds {
--- a/src/typed_pipes.rs
+++ b/src/typed_pipes.rs
@@ -71,6 +71,11 @@ struct BinaryPipeEntry {
     /// capture callback) clone this `Arc` and push frames; the drain thread
     /// pops and writes them to the socket.
     ring: Arc<ArrayQueue<Vec<u8>>>,
+    /// Set by the drain thread before it exits due to a write error (e.g.
+    /// Broken pipe). The host polls this flag per-frame via `drain_failed()`
+    /// and emits an error event so the app is notified promptly rather than
+    /// waiting for the next retry attempt.
+    error_flag: Arc<AtomicBool>,
 }
 
 struct JsonPipeEntry {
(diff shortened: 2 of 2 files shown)


ianjamesburke:
## Test result: ❌ Breaks if condition violated

Tested on the PR build (`plexi-pr-763`, v3.5.4). Two issues found from `~/.plexi-pr-763/plexi.log`:

### 1. Stale session warn still fires

The PR's own `Breaks if` condition is not met — `stale session — cleaning up and restarting` still appears after a broken pipe:

```
[00:09:45] WARN typed_pipes: drain write error: Broken pipe (os error 32)
[00:09:45] WARN process_app: AudioCapture pipe_id=audio-recorder-mic drain failed — cleaning up   ← new path fires correctly
[00:09:58] INFO typed_pipes: drain thread started for pipe audio-recorder-mic
[00:09:58] INFO process_app::routing: audio.capture started
[00:10:00] INFO app::audio-recorder: capture stopped, 0 chunks
[00:10:00] WARN process_app::routing: stale session — cleaning up and restarting   ← old path still reachable
```

Root cause: the `drain_failed` path handles the broken pipe session correctly, but when the app immediately starts a *new* capture, that new session hits the old stale session check (0 chunks reported). The two paths are not mutually exclusive.

### 2. UI freeze after stale session path

Immediately after the stale session warn, the UI thread blocked for 15 seconds:

```
[00:10:01] INFO  heartbeat: beat=1 uptime=00:00:30
[00:10:02] WARN  heartbeat: [FREEZE] UI thread unresponsive for 2s
...
[00:10:15] WARN  heartbeat: [FREEZE] still frozen — 15s unresponsive
```

This is what macOS reported as the hang (incident `AECF671F`, 00:10:15). The `cleaning up and restarting` code path in `stale session` is blocking the UI thread. Even if issue #1 is fixed, this path should be async-safe since it can still be reached in other scenarios.

PR #777, opened 2026-05-06
feat(tts): audio.speak primitive — emit.speak(text) via Gemini TTS

### LINKED ISSUE #747: Host: TTS primitive — emit.speak(text)
## Summary

Add a new PGAP command `speak` that any app can emit to have the host synthesize and play speech audio. Routes to Gemini Flash TTS via direct Google API.

## Requirements

- New PGAP command: `{"cmd": "speak", "text": "Got it", "voice": null}`
- Host-side handler: sends text to Gemini TTS API, receives audio, plays on default output device
- New capability: `audio.speak` — apps must declare this in manifest to use speak
- Python SDK: `emit.speak(text, voice=None)` method on Emitter
- Audio playback uses system default output (CoreAudio on macOS)
- Config in `voice.toml`:
  ```toml
  [voice.tts]
  provider = "google"
  model = "gemini-2.5-flash"
  api_key_env = "GOOGLE_API_KEY"
  ```
- Log: TTS request, latency, audio duration

## Dependencies

- Independent — can be built in parallel with other voice issues
- Will be used by the voice agent daemon (#748) but is a standalone primitive

## Technical Notes

- Direct Gemini API (not OpenRouter) to avoid double-hop latency
- API: `generateContent` with audio output modality, or dedicated TTS endpoint
- Consider streaming playback (start playing before full audio is received) for longer responses
- For MVP: fire-and-forget playback, no interruption/cancel support needed yet

## Acceptance Criteria

- A test app calling `emit.speak("Hello world")` plays audio through speakers
- Capability enforcement: app without `audio.speak` gets denied
- Configurable via voice.toml

## Prior Attempts

**Attempt 1:** PR #777 — implemented `HostCommand::Speak`, `audio.speak` capability, Gemini TTS API call in a background thread (PCM → WAV via hound → rodio playback), Python SDK `emit.speak()`.
**Why it failed:** Host reads the Google API key via `std::env::var("GOOGLE_API_KEY")`, which is not available to GUI apps (Plexi is launched via macOS app bundle, not a shell). Terminal panes inherit the env correctly because Plexi spawns them as login shells — but the host's own Rust threads don't go through that path.
**What to try next:** Replace `std::env::var` with a lookup through the Plexi workspace secrets resolver (the same infrastructure used by `SecretGet` for apps). The host already has a `workspace_root` on the `ProcessApp` — use it to scope the secret lookup. The speak-tester POC app should use `emit.secret_get("GOOGLE_API_KEY")` to surface key availability, not hardcoded static text.

ianjamesburke:
## Summary

- Adds `HostCommand::Speak` to the PGAP protocol — apps emit `{"type":"speak","text":"..."}` and the host handles TTS + playback
- New `audio.speak` capability gates the command; apps must declare it in `manifest.toml`
- Host calls Gemini TTS API in a background thread (non-blocking), decodes PCM → WAV via hound, plays via rodio
- Python SDK: `emit.speak(text, voice=None)` on the `Emitter` class
- Config via `~/.plexi-alpha/config.toml` `[voice.tts]` section (optional — falls back to sensible defaults)
- POC app at `examples/speak-tester/` for manual verification

Closes #747

## Test plan

- [ ] Install PR build: `just pr-install <N>` from feature worktree
- [ ] Install speak-tester app and open it
- [ ] With `GOOGLE_API_KEY` set, press Enter — audio should play through speakers
- [ ] Check `~/.plexi-alpha/plexi.log` for TTS latency and playback log lines
- [ ] Open a second app without `audio.speak` in manifest, call speak — log should show capability denied

### DIFF
--- a/examples/speak-tester/manifest.toml
+++ b/examples/speak-tester/manifest.toml
@@ -1,0 +1,15 @@
+schema_version = 1
+
+[app]
+id = "speak-tester"
+type = "app"
+name = "Speak Tester"
+version = "0.1.0"
+description = "POC app — tests the audio.speak TTS primitive. Press Enter to say 'Hello from Plexi'."
+entry = "speak_tester.py"
+
+[app.capabilities]
+capabilities = ["audio.speak"]
+
+[launch]
+layout_hint = { side = "right", split = 0.3 }
--- a/examples/speak-tester/speak_tester.py
+++ b/examples/speak-tester/speak_tester.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python3
+"""Speak Tester — POC for emit.speak() TTS primitive.
+
+Press Enter to fire a speak command. Watch plexi.log for TTS latency.
+"""
+from plexi_sdk import App, RenderContext
+
+
+class SpeakTester(App):
+    def __init__(self) -> None:
+        super().__init__()
+        self._count = 0
+
+    def on_render(self, ctx: RenderContext) -> None:
+        ctx.text(10, 20, "Speak Tester", size=16, color="#cdd6f4", bold=True)
+        ctx.text(10, 50, f"Press Enter to speak. Count: {self._count}", size=13, color="#a6adc8")
+        ctx.text(10, 75, "Requires GOOGLE_API_KEY to be set.", size=12, color="#6c7086")
+
+    def on_key(self, _ctx: RenderContext, key: str, _mods: dict) -> None:
+        if key == "Return":
+            self._count += 1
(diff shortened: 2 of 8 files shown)


ianjamesburke:
Closing — approach works at the Rust/Python level but the API key lookup is wrong. Host reads GOOGLE_API_KEY via std::env::var which isn't available to GUI apps unless they inherit it from a shell. The right fix is to route through the Plexi workspace secrets store (same as secrets.get for apps). See issue #747 for full context.

PR #797, opened 2026-05-07
fix(context): remove CWD-based auto-switch (#796)

### LINKED ISSUE #796: P0: strip automatic context switching — CWD-based auto-switch is actively disruptive
## What

`check_context_auto_switch()` fires every frame when the focused pane changes. It walks the terminal's CWD up the directory tree looking for a matching context root, and if found silently switches the active context. The ZSH `precmd` hook in the PTY environment emits OSC 7 on every prompt, keeping the CWD fresh for this path.

The feature is not complete and is actively disruptive — users get teleported to a different context simply by navigating directories or focusing a different terminal.

## Reproduce

1. Have two contexts, each with a root path configured
2. Open a terminal in Context A, `cd` into Context B's root (or a subdirectory)
3. Focus away and back, or wait for the next precmd
4. Active context silently switches to Context B — the terminal that was focused disappears from view

## Expected

No automatic context switching occurs. Context switching is always an explicit user action.

## Action Plan

- **`src/app/sync.rs`** — delete `check_context_auto_switch()` entirely and remove its call site in the per-frame update loop
- **`src/app/mod.rs`** — remove fields `last_focused_tile_for_auto_switch: Option<egui_tiles::TileId>` and `skip_auto_switch_frames: u8` from `PlexiApp` and both struct-literal init blocks
- **`src/sidebar.rs:270`** — remove the `skip_auto_switch_frames = 2` assignment (set on manual context switch to suppress auto-switch)
- **`src/command_palette.rs:527`** — same, remove `skip_auto_switch_frames = 2`
- **`src/testing.rs`** — delete `skip_auto_switch_frames_decrements_each_frame` and `auto_switch_no_op_without_context_roots` tests
- Leave the OSC 7 `precmd` hook in `src/shell.rs` intact — it is used for split-pane CWD inheritance, not just auto-switching

See #786 for the longer-term design question of whether and how to re-introduce this as an opt-in feature.

## Done When

`cd`-ing into a directory that matches another context's root (inside any terminal) never triggers an automatic context switch, regardless of focus state.

ianjamesburke:
Closes #796

## What

Removes `check_context_auto_switch()` and all supporting state (`last_focused_tile_for_auto_switch`, `skip_auto_switch_frames`) that silently switched the active context when a terminal's CWD matched another context's root path.

## Why

The auto-switch was actively disruptive — users got teleported to a different context by simply navigating directories or focusing a different terminal pane. The suppression mechanism (`skip_auto_switch_frames=2`) was a race: 2 frames (~33ms) wasn't enough, causing manual context switches to immediately bounce back.

## Changes

- `src/app/sync.rs` — deleted `check_context_auto_switch()` and its call site
- `src/app/mod.rs` — removed `last_focused_tile_for_auto_switch` and `skip_auto_switch_frames` fields
- `src/sidebar.rs` — removed `skip_auto_switch_frames = 2` assignment
- `src/command_palette.rs` — removed `skip_auto_switch_frames = 2` assignment
- `src/testing.rs` — deleted the two tests for this behaviour

The OSC 7 `precmd` hook in `src/shell.rs` is untouched — it's used for split-pane CWD inheritance.

**Breaks if:** Context switches when a terminal `cd`s into another context's root directory.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -264,12 +264,6 @@ pub struct PlexiApp {
     /// Receiver for HostCommands sent over the PLEXI_SOCKET Unix socket listener.
     /// Drained each frame in `drain_pane_cmd_channel`.
     pane_ipc_rx: std::sync::mpsc::Receiver<crate::app_protocol::HostCommand>,
-    /// Last tile that triggered an auto-context-switch check. Avoids re-running
-    /// the CWD lookup every frame when focus hasn't changed.
-    pub(crate) last_focused_tile_for_auto_switch: Option<egui_tiles::TileId>,
-    /// Frames remaining before auto-context-switch is allowed after a manual
-    /// context switch. Prevents immediately overriding the user's intent.
-    pub(crate) skip_auto_switch_frames: u8,
 }
 
 #[cfg(test)]
@@ -628,8 +622,6 @@ impl PlexiApp {
                     update_rx: Some(update_rx),
                     update_available: None,
                     pane_ipc_rx,
-                    last_focused_tile_for_auto_switch: None,
-                    skip_auto_switch_frames: 0,
                 };
             }
         }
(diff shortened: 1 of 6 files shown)


ianjamesburke:
Closing — approach didn't work. See issue #796 for full context.

PR #825, opened 2026-05-08
fix(sidebar): restrict rename double-click to text label hitbox

ianjamesburke:
Fixes #462.

Previously, double-clicking anywhere on a sidebar row triggered rename — including the left drag/grab zone. Now rename only fires when the double-click lands directly on the context name label text.

## Changes

- `sidebar_row.rs`: `draw()` signature updated — `content_fn` now returns `Option<egui::Response>` (the label widget's response). Rename action checks `label_response.double_clicked()` instead of the full-row response. Cursor overrides to `Default` when pointer is over the label rect (was showing `Grab`).
- `sidebar.rs`: Label widget gains `.sense(Sense::click())` and its response is returned from the content closure.

## Breaks if
Rename stops working when double-clicking the context name text in the sidebar.


ianjamesburke:
Closing — approach didn't work as intended. See issue #462 for full context.

PR #908, opened 2026-05-09
fix(input-inspector): add keyboard_capture so Cmd+H/J/K/L reach the app

### LINKED ISSUE #907: fix(input-inspector): add keyboard_capture so Cmd+H/J/K/L reach the app
---
depends_on: []
---

## Problem

`examples/input-inspector/manifest.toml` does not set `keyboard_capture = true`.

Cmd+H/J/K/L are consumed by the host in `poll_actions` for pane navigation and never delivered to the app. For an input inspector whose purpose is to log *all* key events, this is a bug — those events silently disappear.

## Fix

Add `keyboard_capture = true` to `[launch]` in `manifest.toml`.

With keyboard capture active, all host shortcuts except Cmd+Q and Cmd+W are suppressed, so Cmd+H/J/K/L pass through to `on_key`, which already logs any key that isn't `i` or `1–6`.

## Done When

- [ ] Cmd+H/J/K/L pressed while input-inspector is focused appear in the event log

ianjamesburke:
Closes #907

## What

Added `keyboard_capture = true` to `examples/input-inspector/manifest.toml`.

## Why

`poll_actions()` in `src/keys.rs` consumes `Cmd+H/J/K/L` for pane navigation before any app sees them. The input inspector's purpose is to log *all* key events — silently dropping host-consumed shortcuts defeats that. With `keyboard_capture = true` the host suppresses all shortcuts except `Cmd+Q` and `Cmd+W`, so every other combo (including `Cmd+H/J/K/L`) reaches `on_key` and appears in the event log.

**Breaks if:** Cmd+H/J/K/L pressed while input-inspector is focused do not appear as `key` entries in the event log.

### DIFF
--- a/.claude/skills/ship-issue/SKILL.md
+++ b/.claude/skills/ship-issue/SKILL.md
@@ -294,13 +294,18 @@ Reply with one of:
 
 Then send a synchronous notification so the user gets pulled in only when ready to test:
 ```bash
-plexi notify --title "PR #<n> ready to test" \
+RESULT=$(plexi notify --title "PR #<n> ready to test" \
   --body "<title>. Pass criteria: <one-line summary>. Reply pass/fail/modify in pane." \
-  --choice "a:Open PR:open_url:<pr-url>" \
-  --choice "b:Open PR build:open_app:Plexi PR<n>" \
-  --choice "c:Talk to Claude:pane_focus:$SHIP_PANE"
+  --choice "a:Open PR" \
+  --choice "b:Open PR build" \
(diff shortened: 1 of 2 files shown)


ianjamesburke:
Closing — keyboard_capture in manifest is necessary but not sufficient. Root cause: handle_key() in src/process_app/mod.rs:1710 guards against forwarding any Cmd-modified key unconditionally. Even with keyboard_capture=true, Cmd+H/J/K/L are dropped before reaching the app's on_key. Fix must update that guard to allow Cmd-modified keys through when self.keyboard_capture is true (excluding Cmd+Q and Cmd+W).

Issue #563, opened 2026-05-03
idea(navigation): spatial ribbon — niri-style within-context pane ordering, context switching ergonomics, and tabs as spatial positions


ianjamesburke:
---
touches: []
clarification_needed: ["Needs design document before implementation can be scoped."]
---
## Overview

The current tiling model uses a recursive binary split tree navigated by `Cmd+H/J/K/L`, with contexts switched via `Shift+Cmd+H/J/K/L`. This produces two separate navigation systems with no spatial relationship between them. Contexts are represented as opaque dots that are hard to see and rarely used.

This issue describes a unified spatial navigation model inspired by niri (a scrollable-tiling Wayland compositor), i3's named workspace + marks system, and PaperWM — adapted for an agent-management workflow where pane priority is explicit and left-to-right ordering carries meaning.

---

### Within a context: the ribbon

Panes within a context are ordered as a **horizontal priority queue**:

- **Left = highest priority** — the thing you should be doing right now
- **Right = newest / lowest priority** — things appended as new work arrives
- New panes always append to the rightmost position ("back of the ribbon")
- Closing completed work happens from the left as tasks finish
- "I'm lost, what do I do?" = one gesture to jump to the leftmost pane

Each position on the ribbon is a **column**. A column can contain a single pane or a vertical stack of panes (preserving the current vertical split behaviour). Navigation:

- `Cmd+Left / Cmd+Right` — move between columns on the ribbon
- `Cmd+Up / Cmd+Down` — move between panes within a column (vertical stack)

This replaces the four-directional tree navigation (`Cmd+H/J/K/L`) with a two-axis model that maps to a user's actual mental model: horizontal = priority order, vertical = stacked work within one concern.

### Between contexts: named rooms

Contexts remain distinct surfaces (not one infinite canvas). The changes are ergonomic:

- **`Cmd+[` / `Cmd+]`** (or equivalent) — slide between contexts in order, replacing `Shift+Cmd+H/L`
- **Named jump** — type a context name in the command palette to teleport directly; the palette already ranks by recency (#408)
- **Jump to start** — a single shortcut (`Cmd+Shift+Left` or similar) to go to the leftmost pane of the leftmost column in the current context

### Tabs as spatial positions

[...]

PR #954, opened 2026-05-10
feat(navigation): Cmd+Shift+1-9 direct context jump

### LINKED ISSUE #946: idea(navigation): Cmd+Number jump — direct access to context N (Cmd+Shift+N) and lateral pane N (Cmd+N)
---
depends_on: []
---
## What
Numeric jump shortcuts:

- `Cmd+Shift+1–9` — jump directly to context N (0-indexed: Shift+1 → context 0). No-op if context N doesn't exist.
- `Cmd+1–9` — jump directly to lateral pane column N in the current context. **Create the pane if it doesn't exist** (replacing the old `Cmd+Shift+L` lateral split behavior).

### Context jump — `Cmd+Shift+1–9`
- Switches to context index N-1 (1-indexed keys, 0-indexed contexts)
- No-op if context N doesn't exist
- Replaces nothing — new binding

### Lateral pane jump — `Cmd+1–9`
- Jumps to lateral pane column N in the current context
- If column N doesn't exist: **create a new terminal pane** (same behavior as `Cmd+Shift+L` lateral split, now accessible by number)
- Replaces `Cmd+1–9` which previously switched contexts — that binding is now retired in favor of Cmd+Shift+1–9

## Prior Attempts

**Attempt 1 (PR #954):** Implemented Cmd+Shift+1-9 only; changed modifier to Cmd+Option due to macOS screenshot conflict concern.
**Why it failed:** Spec is locked to Cmd+Shift. Also missing the Cmd+1-9 lateral pane behavior entirely.
**What to try next:** Implement both bindings together. Keep Cmd+Shift+1-9 as-is (macOS intercepts Cmd+Shift+3/4 at the system level, but egui may still receive them inside the app window — verify at runtime). Wire Cmd+1-9 to lateral split navigation with create-if-absent.

ianjamesburke:
Closes #946 (context jump only — Cmd+N lateral jump deferred to #563)

## Changes
- `Cmd+Shift+1` through `Cmd+Shift+9` jump directly to context N (0-indexed: Shift+1 → context 0)
- `Cmd+1-9` kept as alias until #563 (spatial ribbon) repurposes it for lateral pane column jumping
- No-op if context N doesn't exist — no wrap, no error
- `log::info!` emitted on every `SwitchContext` dispatch with index and exists flag
- 2 HostHarness tests added: jump to context 1 via Cmd+Shift+2, and no-op when target absent

## Out of scope
`Cmd+1-9` lateral pane jump is blocked on #563 (spatial ribbon) and intentionally excluded.

🤖 Generated with Claude Code

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2182,7 +2182,9 @@ impl eframe::App for PlexiApp {
                     }
                 }
                 Action::SwitchContext(n) => {
-                    if n < self.router.len() {
+                    let exists = n < self.router.len();
+                    log::info!("context jump: Cmd+Shift+{} → context {} (exists: {})", n + 1, n, exists);
+                    if exists {
                         self.switch_workspace(n);
                     }
                 }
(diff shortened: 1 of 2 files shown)


ianjamesburke:
Closing — wrong modifier and incomplete scope.

**What failed:** Modifier changed to Cmd+Option per a macOS screenshot conflict concern, but the locked spec is Cmd+Shift+1-9 for context jump.

**Also missing:** Cmd+1-9 lateral pane navigation (jump to column N in current context, create pane if absent) — this was the old Cmd+Shift+L behavior and must be in the same issue.

**Next attempt:** Reimplement with Cmd+Shift+1-9 for context jump AND Cmd+1-9 for lateral pane navigation (create if absent).

PR #959, opened 2026-05-10
fix(zoom): show pane name bar and tab dots in full-screen overlay

### LINKED ISSUE #861: bug(zoom): full-screen panes don't display pane title section
## What
When a pane is zoomed to full-screen (pressing Enter), the title/name bar that normally appears above the terminal or app content is not displayed. The pane fills the entire screen but lacks the header that shows the pane name and tab indicators.

## Why
The zoom overlay rendering code in `src/app/mod.rs` (lines 2428–2550) directly renders terminal content without calling the proper render function from `src/render/terminal_pane.rs::render()`. The `render_name_bar_and_dots()` function that paints the title bar is skipped entirely in zoom mode.

## Action Plan
- Refactor zoom overlay pane rendering to use or reuse the standard render functions from `src/render/terminal_pane.rs` and `src/render/app_pane.rs`
- Ensure the title/name bar is painted above the pane content (with appropriate top padding/inset)
- Verify tab indicators appear when multiple tabs exist
- Test on both terminal panes and app panes

## Done When
- Running any pane in full-screen (press Enter) shows the pane title bar above the content
- Tab dots appear if the pane has multiple tabs
- Title bar is visible and properly styled in the zoom overlay

## Open Questions
None — the root cause is clear from code inspection.

## Prior Attempts

**Attempt 1:** PR #959 — called `render_name_bar_and_dots` (made `pub(crate)`) inside the `egui::Frame` closure of the zoom overlay, before the `if t.exited` branch. Captured `zoomed_pane_name` from `behavior.pane_names` before `drop(behavior)`. Removed the old `add_space(TAB_DOT_RESERVED_HEIGHT)` and `paint_tab_dots` blocks.

**Why it failed:** Name still doesn't appear in zoom mode per user test. Root cause unknown — `zoomed_pane_name` may be resolving to `None` (lookup miss), or the name bar text is being painted but covered by another layer. No debug logging was included so the exact failure point wasn't confirmed.

**What to try next:**
1. Add `log::warn!` before the `render_name_bar_and_dots` call to confirm `zoomed_pane_name` is non-None and the function is reached.
2. Check whether the Frame's background fill (`self.colors.terminal_bg`) is painted AFTER the name bar text in egui's painter queue — if so, the background would cover the text. If that's the issue, paint the name bar OUTSIDE the Frame (on `child_ui` directly, before `Frame::show`) or use a separate overlay layer (`ui.with_layer_id`).
3. Alternative: bypass the Frame entirely and paint the name bar directly on `child_ui` (the `inner_rect` UI), then create a new child UI with `inner_rect.with_min_y(inner_rect.min.y + 20.0)` for the Frame + terminal.

ianjamesburke:
Fixes #861.

## What

The zoom overlay rendered terminal content directly, bypassing the `render_name_bar_and_dots` call that the normal `pane_ui` path makes. This left the name bar and tab dots invisible whenever a pane was zoomed to full-screen.

## How

- Made `render_name_bar_and_dots` `pub(crate)` so the zoom overlay in `app/mod.rs` can call it directly.
- Capture `zoomed_pane_name` from `behavior.pane_names` before dropping `behavior`, alongside the already-captured `zoomed_tab_info`.
- Call `render_name_bar_and_dots` at the start of the terminal arm (before the exited/hover/normal split), constructing minimal single-entry HashMaps from the captured values.
- Remove the now-redundant `add_space(TAB_DOT_RESERVED_HEIGHT)` and `paint_tab_dots` blocks that were partial stand-ins.

## Breaks if

Named pane in zoom overlay shows no title text, or tab dots disappear from zoomed panes with multiple tabs.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2583,7 +2583,7 @@ impl eframe::App for PlexiApp {
                                                 .map(|n| HashMap::from([(pane_id, n.clone())]))
                                                 .unwrap_or_default();
                                             if zoomed_pane_name.is_some() || zoomed_tab_info.is_some() {
-                                                log::info!(
+                                                log::debug!(
                                                     "zoom: rendering name bar — pane={pane_id:?} name={zoomed_pane_name:?} tabs={zoomed_tab_info:?}"
                                                 );
                                             }
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2520,6 +2520,7 @@ impl eframe::App for PlexiApp {
                         let pane_id = *pane_id;
                         let panel_rect = ui.max_rect();
                         let zoomed_tab_info = behavior.tab_info.get(&zoomed_tile).copied();
+                        let zoomed_pane_name = behavior.pane_names.get(&pane_id).cloned();
 
                         // Drop behavior to release the mutable borrow on ctx.panes
                         drop(behavior);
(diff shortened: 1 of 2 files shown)


ianjamesburke:
Closing — fix didn't work. See issue #861 for failure details.

PR #968, opened 2026-05-10
feat(navigation): Cmd+Shift+1-9 context jump + Cmd+1-9 lateral column jump

### LINKED ISSUE #946: idea(navigation): Cmd+Number jump — direct access to context N (Cmd+Shift+N) and lateral pane N (Cmd+N)

ianjamesburke:
Closes #946

Second attempt — fixes both issues from failed PR #954 (wrong modifier, missing column jump).

## Changes

- **`Cmd+Shift+1-9`** → `SwitchContextDirect(n)`: jumps directly to context N (0-indexed). No-op if N is out of range. Replaces the `Cmd+Option` modifier from the failed attempt.
- **`Cmd+1-9`** → `FocusOrCreateColumn(n)`: focuses the Nth lateral column pane in the current window. Creates a new pane to the right (via `split_focused_mirror`) if column N doesn't exist. Retires the old `Cmd+1-9 → SwitchContext` binding.

## Implementation

**Column lookup** (`Window::column_panes()`): tree-walk from root. At a horizontal linear container, each direct child is a distinct column slot (recurse to find its first visible pane). At any other container (vertical stack, tabs), treat as one column. Returns ordered `TileId` slice without needing rendered rects — works in headless tests.

New `Action` variants: `SwitchContextDirect(usize)`, `FocusOrCreateColumn(usize)`.

## Tests (4 new HostHarness tests)

- `switch_context_direct_in_range_switches` — Cmd+Shift+2 moves to context 1
- `switch_context_direct_out_of_range_noop` — index 8 with 1 context is no-op
- `focus_or_create_column_existing_focuses` — column 0 present → focused_pane set
- `focus_or_create_column_absent_creates_pane` — column 1 absent → spawns terminal pane via IPC, then verifies column_panes() returns 2 and re-focus doesn't create a third

## Verify

```
plexi-pr-<N>
# Cmd+Shift+2 → jumps to context 2 (if it exists)
# Cmd+2 → focuses second column pane or creates one
```

**Breaks if:** `Cmd+Shift+1-9` either does nothing or still switches context via the old path; `Cmd+1-9` switches context instead of focusing/creating a column pane.

🤖 Generated with Claude Code

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2188,6 +2188,15 @@ impl eframe::App for PlexiApp {
                         self.switch_workspace(n);
                     }
                 }
+                Action::SwitchContextDirect(n) => {
+                    log::info!("context jump: Cmd+Shift+{} → context {}", n + 1, n);
+                    if n < self.router.len() {
+                        self.switch_workspace(n);
+                    }
+                }
+                Action::FocusOrCreateColumn(n) => {
+                    self.focus_or_create_column(n);
+                }
                 Action::NextTab => {
                     self.cycle_tab(true);
                 }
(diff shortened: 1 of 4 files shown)


ianjamesburke:
Closing — feature descoped. The problem statement wasn't clear enough to justify the complexity.

PR #977, opened 2026-05-10
feat(create-app): scaffold + open + hot-reload skill (#974)

### LINKED ISSUE #974: feat(skill): create-app — scaffold + open + hot-reload in one Claude Code command
---
depends_on: [992]
---

## Problem

There is no guided path for creating a new Plexi app from scratch. A developer (or Claude Code agent) has to manually scaffold \`manifest.toml\` and \`app.py\`, register the app, open a pane, and wire up hot reload. This should be one command.

## Goal

A Claude Code skill that, when invoked, does all of the following:

1. Prompts for app name and a one-sentence description
2. Scaffolds \`manifest.toml\` + \`main.py\` in a new directory (sensible defaults, minimal working app that renders something visible)
3. Opens a new pane immediately — captures and uses the returned pane ID
4. Enables hot reload for the app directory
5. Reports the new pane ID and working directory so the developer can start editing immediately

## Done When

- [ ] Skill file exists at \`.claude/skills/create-app/SKILL.md\`
- [ ] Running the skill end-to-end opens a live, rendering app in a new pane
- [ ] Hot reload is active — editing \`app.py\` updates the pane without manual restart
- [ ] Skill is self-contained — no undocumented external dependencies

## Notes

This is the intro video workflow: user says "build me an app that does X" → app opens in a pane → agent edits live. The skill is the glue. Prerequisite: \`plexi open\` must return the pane ID to stdout (covered by #937, now closed).

## Context for Next Agent

The `plexi app init` CLI and the host registry both need fixes before this skill can work end-to-end. The SKILL.md itself is straightforward once the plumbing works.

**Key files:**
- `src/cli.rs` — `app_init()` and `scaffold_python_app()` — scaffolding logic
- `src/pane_ops/create.rs` — `launch_app_by_id_with_layout()` — registry lookup on open
- `src/process_app/mod.rs` — PYTHONPATH injection at app launch
- `.claude/skills/create-app/SKILL.md` — the skill deliverable (new file)

## Prior Attempts

**Attempt 1:** PR #977 — feat(create-app): scaffold + open + hot-reload skill

**What was tried:**
1. Added registry rescan on cache miss in `launch_app_by_id_with_layout` — when `plexi open <name>` is called for an app not in the startup registry, rescan from disk. This is the right fix and the log confirms it fires correctly.
2. Fixed `scaffold_python_app` manifest: moved `type = "app"` from root level into `[app]` section (root-level placement silently fails validation with "missing field type").
3. Removed the embedded `PYTHON_SDK` copy from scaffold — the SDK is now a package with relative imports; copying `__init__.py` as a flat file breaks at `from ._constants import`. The host injects PYTHONPATH at launch so no local copy is needed.
4. Added auto-open: when `PLEXI_SOCKET` is set, `plexi app init` calls `open_cli()` immediately after scaffolding so the app opens in one command.
5. Added a `create-app` SKILL.md.

**Why it failed:** Three separate bugs were found and fixed sequentially, but the third `just pr-install` was accidentally run from the repo root instead of the feature worktree — it compiled and installed the alpha binary, not the feature branch. The final test (auto-open not firing) was against unpatched alpha code. The rescan fix was confirmed working in an earlier test run (log line present at 22:59:34).

**What to try next:**
- Start a fresh worktree from the closed PR's branch (`feature/974-create-app-skill`) or cherry-pick its commits onto a new branch
- The three fixes (manifest type placement, SDK copy removal, registry rescan) are all correct — verify they compile and install from INSIDE the feature worktree (`cd worktrees/feature/<branch> && just pr-install <N>`)
- The auto-open uses `crate::cli::open_cli(name, &[], None)` from inside `app_init()` — confirm PLEXI_SOCKET is visible in the CLI subprocess environment (it may not be if the CLI is spawned without inheriting the shell env)
- Use `plexi validate .plexi/apps/<name>` to confirm manifest is valid before testing open

**Attempt 2:** PR #983 — feat(create-app): scaffold + open + hot-reload skill
**Why it failed:** App opens but takes over the full screen — layout_hint (`side = "right", split = 0.5`) in the scaffold manifest is not honored. The rescan-on-miss fix works (app opens), but the layout hint is either ignored by `launch_app_by_id_with_layout` when going through the rescan path, or `plexi open` does not pass the layout hint correctly.
**What to try next:** Before shipping this skill, fix the host to respect `layout_hint` from the manifest when launching via `plexi open`. Grep `launch_app_by_id_with_layout` in `src/pane_ops/create.rs` — the hint is read from the registry after the rescan but may be overridden by a `None` layout arg from `open_cli`. Confirm the hint reaches `open_process_app_pane`.

ianjamesburke:
## Summary

- Adds `.claude/skills/create-app/SKILL.md` — Claude Code skill that scaffolds a Plexi app, patches `watch = true` for hot reload, and opens it in a live pane in one command
- Paired host fix: `launch_app_by_id_with_layout` now rescans the app registry from disk on a cache miss, enabling `plexi open <name>` to work for apps created mid-session via `plexi app init`
- Without the host fix, the skill's `plexi open` step would silently fail because the startup-time in-memory registry never saw the newly created app

## Test plan

- [ ] Diff review: verify `create-app/SKILL.md` covers all 6 steps (validate env, scaffold, patch manifest, open pane, report, edit loop)
- [ ] Diff review: verify host fix in `create.rs` rescans only on cache miss (not on every open)
- [ ] Build: `cargo build` passes ✅

Closes #974

### DIFF
--- a/.claude/skills/create-app/SKILL.md
+++ b/.claude/skills/create-app/SKILL.md
@@ -1,0 +1,136 @@
+---
+name: create-app
+description: Scaffold a new Plexi app, install it, open it in a pane, and activate hot reload — all in one command. Use when the user says "build me an app", "create a Plexi app", or "make an app that does X".
+skill_version: "3.5.88"
+source: local
+date_added: "2026-05-09"
+---
+
+# Create App
+
+Scaffold a minimal working Plexi app, open it in a live pane, and enable hot reload so the agent can edit it immediately.
+
+---
+
+## Step 1 — Gather requirements
+
+Ask (in a single message, not separately):
+
+1. **App name** — a short slug (`snake_case`, no spaces). This becomes the app ID and directory name.
+2. **One-sentence description** — what it does. Used in the manifest and as the initial render hint.
+
+If the user already provided both (e.g. "make an app called `notes` that shows a text editor"), extract them and skip asking.
+
+---
+
+## Step 2 — Validate environment
+
+```bash
+# Must be inside a Plexi pane
+if [ -z "$PLEXI_SOCKET" ]; then
+  echo "error: PLEXI_SOCKET not set — run this inside a Plexi terminal pane"
+  exit 1
+fi
+echo "PLEXI_SOCKET is set: $PLEXI_SOCKET"
+```
+
+```bash
+# Must have a workspace (CWD or ancestor has .plexi/)
+plexi workspace init 2>/dev/null || true   # no-op if already initialised
+```
+
+> If the user is in their home directory (`$HOME`), stop and ask them to `cd` into a project directory first. `plexi workspace init` must never run from `~` — it would collide with the global profile dir.
+
+---
+
+## Step 3 — Scaffold
+
+```bash
+APP_NAME="<name>"
+plexi app init "$APP_NAME"
+```
+
+This creates:
+```
+.plexi/apps/<name>/
+  manifest.toml
+  main.py
+  plexi_sdk.py
+```
+
+---
+
+## Step 4 — Patch manifest
+
+Add `watch = true` to `[app]` so hot reload activates when the pane opens. Also inject the user's description.
+
+Read the generated manifest:
+```bash
+cat ".plexi/apps/$APP_NAME/manifest.toml"
+```
+
+Edit it in-place — use the Read + Edit tools (not sed). The `[app]` section should become:
+
+```toml
+[app]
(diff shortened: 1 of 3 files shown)


ianjamesburke:
Closing — multiple pre-existing bugs in app init pipeline surfaced during testing. Documenting findings on issue #974 for a clean next attempt.

PR #1017, opened 2026-05-10
feat(navigation): Cmd+HJKL falls through to window nav at pane boundary

### LINKED ISSUE #863: ux(navigation): Cmd+HJKL falls through to adjacent window at pane boundary — drop Shift modifier
## What

\`Cmd+Shift+H/J/K/L\` as a separate binding for cross-window navigation is the wrong model. Two modifier levels for the same conceptual action (\"move focus in direction X\") creates unnecessary cognitive overhead. The user shouldn't need to know whether the next pane is inside or outside the current window.

The fix: when \`Cmd+HJKL\` hits a pane boundary with no neighbor in that direction, transparently fall through to \`navigate_page()\` in the same direction. One gesture, one mental model — focus moves until it can't.

## Why

\`Cmd+Shift+H/J/K/L\` only exists because window navigation is wired separately from pane navigation. If the pane navigator checks for a spatial page neighbor on boundary, the Shift modifier becomes redundant. The freed binding slot can be reclaimed or left unbound.

## Action Plan

- **\`src/pane_ops/layout.rs\` — \`navigate()\`** (line 570): When \`find_pane_in_direction_from\` returns \`None\`, call \`self.navigate_page(dx, dy)\` with the equivalent delta for the direction instead of stopping. No edge-pulse for cross-window hops — just silently move.
- **\`src/spatial.rs\` — \`navigate_page()\`** (line 83): Verify it does not create a new window when no neighbor exists in that direction (i.e., silent no-op at the grid boundary). If it does, add an existence check before creating.
- **\`src/keys.rs\`** (lines 162–176): Remove the \`Cmd+Shift+H/J/K/L\` → \`PageLeft/Right/Up/Down\` bindings. Update the comment block at the top of the file.
- **\`src/app/mod.rs\`**: Remove or repurpose \`Action::PageLeft/Right/Up/Down\` — if nothing else emits them after the key removal, delete the variants and their match arms.

## Done When

- \`Cmd+J\` on the bottom pane of a window that has another window below moves focus to the top pane of the window below, with no modifier required.
- \`Cmd+J\` on the bottom pane of the bottommost window is a no-op (no edge pulse, no new window created).
- \`Cmd+Shift+H/J/K/L\` no longer does anything (or is unbound).
- The comment block in \`src/keys.rs\` reflects the new binding set.
- No regression in within-context \`Cmd+HJKL\` pane navigation.

## Relation to Other Issues

\`#563\` (spatial ribbon) would eventually replace this entire navigation model with a different paradigm. This fix improves the current model without blocking that future work.

## Prior Attempts

**Attempt 1:** Removed `PageLeft/Down/Up/Right` actions and `page_*` bindings; `navigate()` falls through to `navigate_page()` at boundary.
**Why it failed:** `consume_key(cmd(), Key::H)` in egui matches any superset of `COMMAND` — so `Cmd+Shift+HJKL` still triggered the navigate action and created new windows.
**What to try next:** Before consuming the navigate key, guard with `!input.modifiers.shift` (or use `input.modifiers == cmd()` exact match) so Shift+Cmd+HJKL is not consumed by the navigate handler.

ianjamesburke:
Closes #863

## What

Removes the separate `Cmd+Shift+H/J/K/L` window navigation bindings. Instead, `Cmd+H/J/K/L` transparently falls through to `navigate_page()` when there is no pane neighbor in the given direction — one gesture, one mental model.

- **`src/pane_ops/layout.rs`** — `navigate()` now calls `navigate_page(dx, dy)` in the `None` branch instead of stopping
- **`src/keys.rs`** — removed `PageLeft/Down/Up/Right` from `Action` enum, `page_*` from `KeyBindings` struct/defaults/overrides/all-bindings/polling
- **`src/config.rs`** — removed `page_*` fields, `overlay_field!` calls, config template comments
- **`src/app/mod.rs`** — removed `Action::Page*` match arms; added 2 regression tests
- **`src/overlays.rs`** — removed Cmd+Shift+HJKL row; updated Cmd+HJKL label to "Move focus (panes & windows)"
- **`src/spatial.rs`** — removed `navigate_or_create_page()` (now dead code)

At the spatial grid boundary, `navigate_page()` is already a pure no-op — so the gesture is silently ignored with no new window created.

## Tests

- `navigate_at_pane_boundary_falls_through_to_page_navigation` — verifies `active_window` switches when at a pane boundary with a window below
- `navigate_at_grid_boundary_is_noop` — verifies no crash/change at the bottommost window

## Breaks if

`Cmd+J` on the bottom pane of a window does not move focus to the window below when one exists in the spatial grid.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2399,18 +2399,6 @@ impl eframe::App for PlexiApp {
                     self.new_context();
                     self.save_workspace();
                 }
-                Action::PageLeft => {
-                    self.navigate_or_create_page(-1, 0);
-                }
-                Action::PageRight => {
-                    self.navigate_or_create_page(1, 0);
-                }
-                Action::PageUp => {
-                    self.navigate_or_create_page(0, -1);
-                }
-                Action::PageDown => {
-                    self.navigate_or_create_page(0, 1);
-                }
                 Action::ToggleMinimap => {
                     self.minimap.toggle();
                 }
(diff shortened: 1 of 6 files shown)


ianjamesburke:
Closing after first test pass. Bug: `Cmd+Shift+H/J/K/L` still creates new windows. Root cause: `consume_key(cmd(), Key::H)` in egui matches any superset of `COMMAND` modifiers, so the navigate action fires on `Cmd+Shift+HJKL` too. The fix needs to explicitly guard against Shift being held before dispatching the navigate action (e.g. check `!input.modifiers.shift` before consuming). Relabeling as ready for a clean attempt.

PR #1027, opened 2026-05-11
fix(keybindings): Cmd+R rename-pane and font-size bindings broken by exact-modifier guard

### LINKED ISSUE #1021: bug(keybindings): Cmd+R rename-pane broken since #985 — spurious exact-modifier guard
## What

Cmd+R no longer triggers pane rename. The binding is defined and the dispatch branch exists, but the action is never fired.

## Reproduce

1. Launch alpha build (v3.5.111+, after #985 merged)
2. Press **Cmd+R** with a pane focused
3. Expected: rename modal opens
4. Actual: nothing happens

## Root Cause

PR #985 (user-configurable keybindings) introduced an extra exact-modifier guard for `rename_pane` not present on any other binding. In `src/keys.rs` lines 545–548:

\`\`\`rust
} else if input.modifiers == bindings.rename_pane.0
    && input.consume_key(bindings.rename_pane.0, bindings.rename_pane.1)
{
    actions.push(Action::RenamePane);
}
\`\`\`

The commit message for #985 explicitly names this: *"Use exact modifier equality for rename_pane and font-size guards"* (CodeRabbit review change). No other binding uses this pattern — `rename_context` immediately above uses plain `consume_key(mods, key)`. The extra `input.modifiers == bindings.rename_pane.0` equality check likely fails due to how egui normalises macOS modifier flags (`mac_cmd` vs `command`), causing the branch to never fire.

## Action Plan

- `src/keys.rs` lines 545–548: remove the `input.modifiers == bindings.rename_pane.0 &&` guard, leaving just `consume_key` — matching the pattern of `rename_context` and every other binding in the file
- Verify the font-size guard mentioned in the same commit message (`Cmd+=` / `Cmd+-`, issue #1019) — it may have the same problem

## Done When

Pressing **Cmd+R** with a pane focused opens the rename modal in the alpha build.

ianjamesburke:
## Summary

Removes the spurious `input.modifiers == <binding>.0 &&` equality check that was added in #985 for `rename_pane`, `increase_font_size`, and `decrease_font_size`. The check fails on macOS due to egui normalising `mac_cmd` vs `command` differently, so these bindings never fired. Every other binding uses plain `consume_key` — this restores consistency.

**Files changed:** `src/keys.rs` — 3 lines removed (one per affected binding)

## Test plan
- [ ] Press **Cmd+R** with a pane focused → rename modal opens
- [ ] Press **Cmd+=** → font size increases
- [ ] Press **Cmd+-** → font size decreases
- [ ] Cmd+Shift+R still opens rename context

Fixes #1021

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -542,7 +542,7 @@ pub fn poll_actions(
         // Rename context before rename pane — check shifted variant first.
         if input.consume_key(bindings.rename_context.0, bindings.rename_context.1) {
             actions.push(Action::RenameContext);
-        } else if input.consume_key(bindings.rename_pane.0, bindings.rename_pane.1) {
+        } else if !input.modifiers.alt && input.consume_key(bindings.rename_pane.0, bindings.rename_pane.1) {
             actions.push(Action::RenamePane);
         }
 
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -542,9 +542,7 @@ pub fn poll_actions(
         // Rename context before rename pane — check shifted variant first.
         if input.consume_key(bindings.rename_context.0, bindings.rename_context.1) {
             actions.push(Action::RenameContext);
-        } else if input.modifiers == bindings.rename_pane.0
-            && input.consume_key(bindings.rename_pane.0, bindings.rename_pane.1)
-        {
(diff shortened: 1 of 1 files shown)


ianjamesburke:
Closed by PR #1027 — verified on alpha v3.5.114

PR #1086, opened 2026-05-11
fix(notify): anchor normal notifications to top-right toast position

### LINKED ISSUE #1084: fix(app-init): remove ctx.notify() call from default app scaffold
## What

The default app generated by \`plexi app init\` calls \`ctx.notify()\` on every button click, firing a numbered notification badge (🔔 N) in the Plexi toolbar. This is confusing for new users — it looks like a bug and clutters the UI out of the box.

## Location

\`src/cli.rs\` — the scaffolded app template. Look for \`ctx.notify()\` in the default \`on_click\` / button handler block (around line 679).

## Done When

\`plexi app init\` generates an app with no \`ctx.notify()\` call in the default scaffold. Clicking the default button does not trigger a notification or badge.

The \`ctx.notify\` API can remain in a comment as an example, but must not fire on every click by default.

ianjamesburke:
## What

Notification modal was anchored at `CENTER_CENTER` of the full window, overlapping app pane content.

## Changes

- `PRIORITY_CRITICAL` (≥200): unchanged — CENTER_CENTER + full-window scrim forces user attention
- All other priorities (HIGH, NORMAL, LOW): anchored to `RIGHT_TOP` with `-16px` right margin and `80px` top offset to clear the header bar; no scrim so app remains interactive

## Testing

Trigger a `PRIORITY_NORMAL` notification from any app. The modal should appear in the top-right corner, not covering the app content.

Closes #1084

### DIFF
--- a/.claude/skills/ship-issue/SKILL.md
+++ b/.claude/skills/ship-issue/SKILL.md
@@ -155,6 +155,18 @@ Before writing any code, read the issue and the relevant codebase to produce a t
 
 **When prior commits exist on the branch:** Before reading worktree files, grep the issue's key identifiers (field names, function names listed in "Files to change" or "Done when") directly on alpha (`src/` in the repo root). If a grep hit confirms an identifier already exists on alpha, that criterion is already implemented — skip reading its worktree file and focus Phase 3 only on what's genuinely missing. This avoids reading files that are identical to alpha and misreading branch-vs-alpha diffs.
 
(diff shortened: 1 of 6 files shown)


ianjamesburke:
Closing — wrong fix. Issue #1084 is actually about removing the notification badge from the default app generated by `plexi app init` (in cli.rs), not the Plexi notification modal positioning.

PR #1307, opened 2026-05-14
feat(inspector): inline context name editing from inspector modal

### LINKED ISSUE #1289: feat(inspector): inline context name editing from inspector modal
## What

The context inspector modal currently displays the context name as a read-only label. It should allow inline renaming — clicking or pressing a key on the context name should activate an editable text field, reusing the existing `rename_buffer` pattern from `draw_rename_context_overlay`.

## Why

Users expect to be able to rename a context from its detail view. Currently they must close the inspector and use the sidebar or Cmd+Shift+R shortcut. Inline editing reduces friction.

## Action Plan

- **`src/overlays.rs:1531–1536`** — Replace the static `RichText::new(&ctx_name)` label with a clickable/hoverable element. On click (or a designated key like `R`), switch to an inline `TextEdit::singleline` using `self.rename_buffer`, matching the pattern in `draw_rename_context_overlay` (`src/overlays.rs:639–710`).
- **`src/overlays.rs:1430`** — Consume `R` key in the input block to trigger rename mode (only when not already editing).
- **`src/app/mod.rs`** — Add `inspector_renaming: bool` field to `PlexiApp` to track whether the inspector is in rename mode. Initialize to `false` in all constructors.
- On Enter/Escape or focus loss: commit the rename via `self.router.active_mut().name = new_name` (same as `draw_rename_context_overlay`), then exit rename mode.

**Future: context description editing** — once #1120 (context descriptions) lands, add a second editable field below the name for the description. This issue does not depend on #1120; name editing ships independently.

## Done When

- Pressing `R` (or clicking the context name) in the inspector switches the name to an editable text field
- Enter commits the rename; Escape cancels
- The rename persists (same behavior as Cmd+Shift+R rename overlay)
- Context description editing is deferred to a follow-up after #1120

## Open Questions

None — the rename_buffer pattern is well-established and can be reused directly.

ianjamesburke:
## Summary

- Adds inline rename capability to the context inspector modal — press `R` or click the context name to edit it in place
- Enter commits the rename (persists via `save_workspace`), Escape cancels
- Reuses existing `rename_buffer` + `TextEdit` pattern from `draw_rename_context_overlay`
- Footer key hints update contextually (rename mode shows Enter/Esc; normal mode shows R/j/k/Enter)

## Test plan

- [ ] Open inspector (`Cmd+Shift+I` or however it's bound)
- [ ] Press `R` — context name becomes an editable text field, pre-selected
- [ ] Type a new name, press Enter — name updates, inspector stays open
- [ ] Press `R` again, then Escape — reverts to original name
- [ ] Click the context name — also enters rename mode
- [ ] Verify footer hints change between rename and normal mode
- [ ] Close and reopen inspector — renamed context persists

Closes #1289

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -193,6 +193,7 @@ pub struct PlexiApp {
     pub(crate) pending_close: bool,
     pub(crate) show_context_inspector: bool,
     pub(crate) inspector_selected_pane: usize,
+    pub(crate) inspector_renaming: bool,
     pub(crate) welcome_delete_press_count: u8,
     pub(crate) welcome_delete_last_press: Option<std::time::Instant>,
     pub(crate) frame_tick: crate::logging::FrameTick,
@@ -663,6 +664,7 @@ impl PlexiApp {
                     quit_last_press: None,
                     show_context_inspector: false,
                     inspector_selected_pane: 0,
+                    inspector_renaming: false,
                     welcome_delete_press_count: 0,
                     welcome_delete_last_press: None,
                     config: config.clone(),
@@ -776,6 +778,7 @@ impl PlexiApp {
             quit_last_press: None,
             show_context_inspector: false,
             inspector_selected_pane: 0,
+            inspector_renaming: false,
             welcome_delete_press_count: 0,
             welcome_delete_last_press: None,
             config,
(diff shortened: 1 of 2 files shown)


ianjamesburke:
Focus fight: text field rapidly focuses and unfocuses when entering rename mode. Same pattern seen in other PRs — likely an upstream issue with cursor focus handling in modals. Not a bug in this PR's logic, but in how the host manages focus layers for text inputs inside overlays.

Issue #1282, opened 2026-05-13
ux(scratchpad): make double-space editor feel shell-native, not app-overlay


ianjamesburke:
## Problem

The scratchpad (src/scratchpad.rs) opens as a host-builtin egui overlay. PR #1412 proved the egui approach is a dead end: focus bugs, host shortcuts don't pass through, feels like a mode switch not a shell feature.

## Solution

Replace the egui scratchpad with an embedded terminal editor. Shift+Cmd+Space spawns `micro` (or similar single-binary TUI editor) as a subprocess in the current pane's PTY. The terminal never stops being a terminal, so focus, keybindings, and rendering all come free.

## Action Plan

- [ ] Replace double-space trigger with Shift+Cmd+Space keybinding
- [ ] On trigger: write current scratchpad content (if any) to a temp file, spawn `micro <tmpfile>` as a subprocess in the focused pane
- [ ] On editor exit: read the temp file contents back as the scratchpad value
- [ ] Remove `src/scratchpad.rs` egui overlay and `open_builtin_app_pane()` overlay path
- [ ] Bundle `micro` binary or document it as a dependency

## Done When

- Shift+Cmd+Space opens a real terminal editor in the current pane
- Editing feels shell-native (same font, same cursor, same keybindings)
- No focus bugs on Cmd+Tab away and back
- Save/quit returns to normal shell with content captured

PR #1412, opened 2026-05-17
fix(scratchpad): re-request focus on window regain after Cmd+Tab (#1381)

### LINKED ISSUE #1381: fix(ui): text editor loses focus when switching away and back
## What

When the text editor is open and you switch focus away from Plexi (e.g. Cmd+Tab to another app) then return, the text editor loses focus. You have to click or re-trigger it to continue typing.

## Action Plan

- [ ] Investigate focus restoration on window regain — the text editor should re-acquire focus if it was the active input when focus was lost
- [ ] Check `eframe`/`egui` focus handling on window activation events

## Done When

- Switch away from Plexi and back: text editor retains focus and accepts keystrokes immediately

ianjamesburke:
## Summary

- The `focus_requested` flag in `ScratchpadApp` is a one-shot guard: set to `true` on the first render, never reset. When the OS window loses focus (Cmd+Tab) and regains it, egui clears its internal focus state for all text widgets, but `focus_requested` stays `true` — so no re-request fires and the text editor is left un-focused.
- Fix: listen for `egui::Event::WindowFocused(true)` in both `Mode::Editing` (main text area) and `Mode::SaveModal` (filename field). When the event fires, `request_focus` is called for that frame's active field.
- Adds `log::info!` trace when the regain path is taken.

## Done When

- Switch away from Plexi while Scratchpad (Cmd+Shift+Space) is open, then switch back — text editor retains focus and accepts keystrokes immediately.

## Test plan

- [ ] Open Scratchpad (Cmd+Shift+Space)
- [ ] Type some text to confirm focus is working
- [ ] Cmd+Tab to another app and back
- [ ] Type immediately — keystrokes should appear in the editor without clicking

Closes #1381

### DIFF
--- a/src/scratchpad.rs
+++ b/src/scratchpad.rs
@@ -123,9 +123,15 @@ impl App for ScratchpadApp {
             );
 
             let te_id = egui::Id::new("scratchpad_text");
-            if !self.focus_requested {
+            let window_regained_focus = ui.ctx().input(|i| {
+                i.events.iter().any(|e| matches!(e, egui::Event::WindowFocused(true)))
+            });
+            if !self.focus_requested || window_regained_focus {
                 ui.ctx().memory_mut(|m| m.request_focus(te_id));
                 self.focus_requested = true;
+                if window_regained_focus {
+                    log::info!("scratchpad: window regained focus — re-requesting text editor focus");
+                }
             }
 
             ui.allocate_new_ui(egui::UiBuilder::new().max_rect(text_rect), |ui| {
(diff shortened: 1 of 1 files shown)


ianjamesburke:
Closing — approach didn't work. The egui TextEdit correctly regains focus on Cmd+Tab, but host split-key shortcuts still don't pass through while the scratchpad captures keyboard input. Fixing focus in the builtin app is the wrong layer; the whole scratchpad needs to be replaced with a terminal-based editor so it behaves like a normal pane. Superseded by #1282.

PR #1417, opened 2026-05-17
fix(inspector/sidebar): editable root path in inspector + sidebar fallback for empty contexts (#1415)

### LINKED ISSUE #1415: bug(inspector/sidebar): no UI path to set context root when context has no panes
## Summary

When a context has no open panes (welcome screen), there is no UI gesture to set its root path — the right-click option requires a focused pane CWD, and the Cmd+I inspector is read-only. This creates a broken loop: you need a pane to set the root, but without the root set new panes open at the wrong directory.

## What

Two compounding gaps:

**1. Sidebar right-click hides \"Set root\" when no panes are open**

\`focused_cwd\` in \`sidebar.rs:70-72\` is derived from \`active_window\`'s focused pane. When the context is empty, \`focused_cwd = None\` and \`cwd_for_menu = None\`, so the \"Set root to current path\" button (line 219-225) is conditionally hidden. The right-click menu shows nothing in the \"Set root\" slot for an empty context.

**2. Inspector modal (Cmd+I) shows root as read-only**

\`render_inspector_header\` (overlays.rs:159-182) renders the root path as a dim label with a copy button — no text input, no path picker, no set/clear controls. There is no keyboard path to set or change the context root.

**Consequential: first pane from welcome screen may open at wrong CWD**

\`cwd_for_welcome_tab\` (layout.rs:1203-1205) correctly consults \`router.active().root\` — the #1351 fix is present in code. But because the root can't be set without panes, the fix is unreachable for new or freshly-cleared contexts. The pane falls back to \`window.path\` (launch CWD), which is typically \`~\` or \`/\`.

**Related: empty contexts don't appear as groups in the inspector**

\`collect_inspector_rows\` (overlays.rs:1631) skips contexts with no panes (\`if !rows.is_empty()\`). Non-active empty contexts are invisible in the inspector entirely.

## Reproduce

1. Create a new context (it starts empty, showing the welcome screen)
2. Right-click the context in the sidebar → no \"Set root\" option appears
3. Press Cmd+I → inspector shows context name + \"No panes\", no way to set root
4. Press ⌘N to open the first terminal — it opens at \`~\` or \`/\`, not any project directory
5. The only workaround is: open a terminal first, \`cd\` to the target dir, then right-click to set root — backwards from intended UX

## Expected

- The inspector modal (Cmd+I) has a text input or path picker to set/clear the context root, accessible even when no panes are open
- Empty contexts appear in the inspector so they can be configured without opening a pane first
- OR: the sidebar right-click offers a \"Set root...\" option that opens a path input dialog, regardless of whether panes are open

## Action Plan

- [ ] \`src/overlays.rs: render_inspector_header\` — add a text input (or editable label) for the root path. When the user types a valid path and confirms (Enter), emit a \`SetContextRoot\` host event. Show a \"Set root\" placeholder when root is \`None\`.
- [ ] \`src/overlays.rs: draw_context_inspector\` — wire the root edit action to \`self.router.active_mut().root = …\` + \`self.save_workspace()\`
- [ ] \`src/overlays.rs: collect_inspector_rows\` (line 1631) — include empty contexts in the groups list so they appear in the inspector (with a \"No panes\" placeholder row rather than being skipped entirely)
- [ ] \`src/sidebar.rs\` (line 219-225) — show a \"Set root...\" option even when \`cwd_for_menu\` is None; clicking it should open a text-input dialog or delegate to the inspector

## Done When

- Pressing Cmd+I on an empty context shows the context name + an editable root path field
- Typing a path and confirming sets the context root without opening a pane first
- Pressing ⌘N from the welcome screen of that context opens the terminal at the newly set root
- Empty contexts appear in the inspector's context list

## Prior Attempts

Two related issues were closed but did not address the UX gap:
- **#1299** (fix: new panes/tabs use context root directory) — fixed the fallback chain in pane creation paths
- **#1351** (fix: first pane from welcome screen falls back to context path) — fixed \`cwd_for_welcome_tab\` to consult \`context.root\`

Neither addressed the missing UI to set the root in the first place. Both fixes work correctly in code but are unreachable when a context starts empty.

ianjamesburke:
## Summary

Breaks the chicken-and-egg loop where setting a context root required an open pane, but the root was needed to open panes at the correct CWD.

## Changes

**Inspector modal (Cmd+I):**
- Root path section is now editable — pencil button (✎) when root is set, "Set root…" button when none
- Enter commits the path (empty = clear root), Escape cancels without dismissing the inspector
- Root edit ESC is consumed before the inspector's global ESC handler, so cancelling edit doesn't close the modal

**Sidebar right-click:**
- Empty contexts (no focused pane) now show "Set root…" instead of hiding the root section entirely
- Clicking it switches to that context, opens the inspector, and pre-activates root edit mode

## Done When (from issue)

- [x] Pressing Cmd+I on an empty context shows an editable root path field
- [x] Typing a path and confirming (Enter) sets the root without opening a pane first
- [x] Empty context right-click shows "Set root…" option
- [x] Pressing ⌘N from the welcome screen of that context opens the terminal at the newly set root (via existing `cwd_for_welcome_tab` which already reads `router.active().root`)

Closes #1415


ianjamesburke:
Closing — approach was too narrow. The root path editing needs a proper keyboard-accessible path input primitive. Superseded by a broader context system rework.

PR #1453, opened 2026-05-18
feat(contexts): remove depth-3 hard cap, allow unlimited nesting (#1392)

### LINKED ISSUE #1392: feat(contexts): remove depth-3 hard cap, allow unlimited nesting with soft warning
---
depends_on: []
---

## What

Commit to the **portal model** for sub-contexts and remove the depth-3 cap. A SubContext tile is a portal — a teleport target that always inserts cleanly as a normal split — not an inline replacement that hijacks the parent's focused pane and leaves a layout gap.

The visual upgrade (live miniatures inside SubContext tiles) is tracked separately in #1495. This issue is the structural fix.

## The keystone change

**Kill `new_child_context`'s pane-adoption branch.** Today it removes the parent's focused pane and replaces the tile in-place with a SubContext, which leaves a layout gap and a "large blank black panel" at depth 2+. Replace with: child contexts always start fresh with their own terminal, and the parent gets a SubContext tile inserted via the standard split path (same primitives as any other pane).

This is what makes \"portal\" real at the layout level: the parent's existing panes stay put, and the SubContext is just another tile in the same tree — no in-place mutation, no orphaned slots, no special case at depth N.

## Why

The portal metaphor only works if portals are normal tiles. Adoption was a clever shortcut that broke recursion: at depth 2+ the adopted pane leaves a gap the SubContext tile doesn't fill, producing the blank-panel bug. Stop fighting it — delete it. With a clean insert path, the same primitive works at every depth and the cap becomes meaningless.

## Action Plan

- [ ] **Restore** PR #1470's `delete_context()` cascade + `depth_stack` cleanup + 3 HostHarness tests (PR #1470 was closed unmerged — re-apply the fix). Cascade: deleting a parent context recursively deletes all descendant contexts. Stack: deleting a context that appears in `router.depth_stack` removes the matching entries.
- [ ] **`new_child_context`** (`src/pane_ops/workspace.rs:14`): delete the adoption branch entirely. Always create a fresh terminal in the child window. Always insert the SubContext tile in the parent via the standard split path (same code path as a normal pane split). No in-place tile mutation, no special-case "if no adoptable pane" branch.
- [ ] **Remove the depth-3 cap** in `new_child_context` (`src/pane_ops/workspace.rs:20`).
- [ ] **Breadcrumb truncation** (`src/sidebar.rs:42–62`): at depth 5+ show first + … + last 2. Dim breadcrumb labels at depth 3+ using `colors.text_dim`.
- [ ] **HostHarness test**: create a depth-4 chain, verify no panics, verify each parent window has a SubContext tile that points at its child and renders without leaving a gap.
- [ ] **HostHarness test**: cascade delete — create parent → child → grandchild, delete parent, verify grandchild + child contexts are both removed and `depth_stack` is empty.

## Constraints

- **SubContext tile rendering stays as text tiers** (existing `sub_context_responsive_tiers`). Live miniatures are #1495 — explicitly out of scope.
- **Sidebar stays flat.** Sub-contexts do not appear in the sidebar; they appear only as portals in their parent and in the breadcrumb. Do not regress.
- **CLI `context new --parent` still works** and goes through the same `new_child_context` after the adoption branch is removed.

## Done When

1. Sub-contexts at depth 4+ work with no blank panels.
2. Adoption branch is gone — child contexts start with their own fresh terminal.
3. Creating a sub-context inserts a SubContext tile via the standard split path; the parent's existing panes are not touched.
4. Deleting a parent context recursively deletes all descendants and cleans up `depth_stack` (3 HostHarness tests green).
5. Breadcrumb truncates at depth 5+ and dims at depth 3+.
6. Depth-4 chain test passes; cascade-delete test passes.

## Prior Attempts

**Attempt 1:** PR #1453 — removed depth cap, added breadcrumb truncation and dim indicator.
**Why it failed:** `delete_context()` didn't clean up `depth_stack`. Breadcrumb rendered \`? › hi\`.

**Attempt 2:** PR #1470 — fixed `delete_context()` with cascade delete + `depth_stack` cleanup + 3 HostHarness tests, then removed the depth cap.
**Why it failed:** Adoption branch left a blank black panel at depth 2+. The delete-side fix was correct but the layout architecture itself (adopt-and-replace-in-place) was unsound.

**Attempt 3 (this issue):** Stop fighting the adoption branch — delete it. Child contexts always start fresh. SubContext tiles insert via the standard split path. Live-miniature rendering split into #1495 so this PR stays focused on the structural fix.

ianjamesburke:
## Summary

- Removes the hard depth-3 cap in `new_child_context()` — users can now create sub-contexts at any depth
- Replaces the blocking `log::warn!` + early return with a non-blocking `log::info!` advisory at depth >= 3
- Breadcrumb bar truncates gracefully at depth 5+ (first + `…` + last 2 segments)
- Dim "deep" label appears in breadcrumb bar at depth >= 3 as a soft nudge

## Done When

- [x] Can create sub-contexts at depth 5, 10, whatever
- [x] Breadcrumb bar doesn't overflow or break at deep nesting
- [x] Soft warning appears at depth 3 but doesn't block creation

Closes #1392

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -32,7 +32,7 @@ pub struct Context {
     pub context_id: u64,
     /// Parent context_id for sub-contexts. None = top-level.
     pub parent_id: Option<u64>,
-    /// Nesting depth. 0 = root level. Capped at 3.
+    /// Nesting depth. 0 = root level. No hard cap; a soft warning is emitted at depth 3+.
     pub depth: u32,
 }
 
--- a/src/pane_ops/workspace.rs
+++ b/src/pane_ops/workspace.rs
@@ -10,7 +10,7 @@ use std::path::PathBuf;
 impl PlexiApp {
     /// Create a child context nested inside `parent_name`. Moves the parent's focused
     /// pane into the new child context and replaces it with a SubContext tile. Falls
-    /// back to a new terminal if no adoptable pane is focused. Depth is capped at 3.
+    /// back to a new terminal if no adoptable pane is focused. No hard depth limit.
     pub(crate) fn new_child_context(&mut self, parent_name: &str, path: PathBuf) -> Result<(), String> {
         let parent_idx = self.router.position(|c| c.name == parent_name)
             .ok_or_else(|| format!("no context named '{parent_name}'"))?;
@@ -18,14 +18,10 @@ impl PlexiApp {
         let parent_depth = self.router.get(parent_idx).depth;
 
         if parent_depth >= 3 {
-            log::warn!(
-                "new_child_context: depth limit reached — parent '{}' is at depth {}",
+            log::info!(
+                "new_child_context: deep nesting — parent '{}' is at depth {} (consider reorganizing)",
                 parent_name, parent_depth
             );
-            return Err(format!(
-                "depth limit: parent '{}' is already at depth {} (max 3)",
-                parent_name, parent_depth
-            ));
         }
 
         let child_depth = parent_depth + 1;
--- a/src/sidebar.rs
+++ b/src/sidebar.rs
@@ -52,15 +52,47 @@ impl PlexiApp {
                     .collect();
                 path_ids.push(self.router.active().context_id);
 
-                for (i, &ctx_id) in path_ids.iter().enumerate() {
-                    let name = self.router.iter()
-                        .find(|c| c.context_id == ctx_id)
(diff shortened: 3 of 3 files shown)


ianjamesburke:
Closing — approach didn't work. See issue #1392 for full context.

PR #1497, opened 2026-05-18
fix(file-browser): vim letter keys now append to query in search mode (#258)

### LINKED ISSUE #258: File browser: stability and UX polish
General tracking issue for file browser bugs and UX rough edges noticed during daily use on alpha.

## Known symptoms
- Unspecified navigation/interaction bugs reported during v2 development (2026-04-15 session)

## When to fix
Not blocking v2. Pick up during a quiet period or when a specific symptom is worth a focused fix. Add concrete repro steps as you find them — use the new debug logging (`level = "debug"` in config, then `grep file_browser ~/.plexi-alpha/plexi.log`) to capture exact key sequences.

ianjamesburke:
## Summary

- `h` in search mode was navigating to the parent directory instead of appending to the query
- `l` in search mode was activating the selected entry instead of appending to the query
- `s` and `r` were toggling sort and refreshing instead of appending
- `/` could not be typed in a search query (it re-triggered EnterSearch)

**Root cause:** `classify_key` was mode-agnostic and returned early on vim key matches before reaching the `AppendText` fallback. Fix adds `in_search: bool` param and gates `H`, `L`, `S`, `R`, and `/` behind `!in_search`.

**Also:** Derives `Debug` for `SortMode` (needed for `assert_eq!` in new regression tests).

## Done when

- 6 new regression tests pass (all 25 file_browser tests green)
- `h`, `l`, `s`, `r` append to search query in search mode
- `←` (ArrowLeft) still navigates up in search mode (arrow keys unchanged)
- `j`/`k` still navigate the filtered results in search mode (unchanged)

Closes #258

### DIFF
--- a/src/file_browser/helpers.rs
+++ b/src/file_browser/helpers.rs
@@ -59,7 +59,7 @@ pub(crate) struct Entry {
     pub modified: Option<SystemTime>,
 }
 
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum SortMode {
     RecentlyTouched,
     Name,
--- a/src/file_browser/mod.rs
+++ b/src/file_browser/mod.rs
@@ -46,7 +46,9 @@ fn key_pressed_no_repeat(input: &egui::InputState, key: egui::Key) -> bool {
     })
 }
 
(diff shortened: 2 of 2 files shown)


ianjamesburke:
Closing — two bugs found during testing. See issue #258 for full context.

PR #1563, opened 2026-05-19
fix(cli): plexi open exits 0 with pane ID when app launch silently fails (#1542)

### LINKED ISSUE #1542: fix(cli): plexi open exits 0 with pane ID when app launch silently fails
## Summary
\`plexi open\` returns a pane ID and exits 0 even when the app fails to launch — capability errors are logged but never reach the caller.

## What
When \`launch_app_by_id_with_layout\` fails (e.g. static capability mismatch, app not found), it logs the error and returns \`()\`. The \`spawn_pane\` IPC handler unconditionally writes \`{\"pane_id\": N}\` to the response file regardless — the pane ID was pre-allocated before the launch attempt. \`open_cli\` reads \`pane_id\`, prints it, and exits 0. The user sees a number but no pane appears, with no indication of what went wrong.

## Reproduce
```
plexi-pr-1538 open webcam-viewer --layout split_h
# Prints: 2
# No pane appears
# ~/.plexi-pr-1538/plexi.log shows:
# [ERROR] ProcessApp[webcam-viewer]: static capability validation failed:
# App declares no 'video.playback' capability but calls emit.open_video().
```

## Root Cause (verified)
- `src/pane_ops/create.rs:1484`: `new_pane_id = self.host.next_pane_id()` allocated before launch attempt
- `src/pane_ops/create.rs:575`: `launch_app_by_id_with_layout` returns `()` — no success/failure signal to caller
- `src/app/mod.rs:1540–1544`: response file written unconditionally as `{\"pane_id\": N}` after launch (success or failure)
- `src/cli.rs:2856–2858`: `open_cli` checks for `pane_id`, prints it, returns 0 — no `error` field check

Same silent path exists in `app_run_cli` (~line 1210) which uses the same response file pattern.

## Action Plan

**`src/pane_ops/create.rs`**
- Change `launch_app_by_id_with_layout` return type from `()` to `Result<(), String>`
- Return `Ok(())` on all success paths
- At line 665 (not found / launch failed): replace `log::warn!` + implicit return with `return Err(message)` — keep the warn log, add the Err
- Update `launch_app_by_id` wrapper to call `.ok()` or propagate (no callers outside IPC and internal use need the error today)

**`src/app/mod.rs`** (SpawnPane IPC handler, ~line 1527)
- Capture result: `let launch_result = self.launch_app_by_id_with_layout(...)`
- On `Err(msg)`: write `{\"error\": \"<msg>\"}` to response file (skip writing pane_id)
- On `Ok(())`: write `{\"pane_id\": new_pane_id}` as today
- Same pattern for the `launch_app_by_path_with_layout` branch (~line 1525)

**`src/cli.rs`**
- `open_cli` (~line 2855): after parsing response JSON, check `error` field before `pane_id`; on error: `eprintln!(\"error: {msg}\")`, return 1
- `app_run_cli` (~line 1210): same check — currently only looks for `pane_id`

## Done When
1. `plexi open webcam-viewer` when webcam-viewer has a capability mismatch prints the error to stderr and exits 1 — no pane ID printed.
2. `plexi open webcam-viewer` when the app launches correctly still prints the pane ID and exits 0.
3. `plexi app run <path>` follows the same contract.
4. The log ERROR line is unchanged (logging is not affected, only response propagation).

## Notes
- `plexi app init` already scaffolds a `manifest.toml` with `capabilities = []` — that's correct and intentional. The dev is expected to fill in capabilities as they use SDK calls. The gap is that mismatches were previously invisible at the CLI boundary.
- A `plexi app validate <name>` command (exposing `static_capability_check` directly) is a natural follow-on — file separately. See also #1346 (`plexi doctor`) for batch auditing.

## Prior Attempts

**Attempt 1:** PR #1563.
**Why it failed:** Host-side response file correctly writes `{"error":"..."}` format. CLI-side error check in `open_cli` (src/cli.rs) does not trigger — raw JSON printed to stdout with exit 0 instead of formatted error to stderr with exit 1. Root cause unconfirmed: either the `open_cli` code path isn't reached, or there's a fallback that prints the raw response content before the error check fires.
**What to try next:** Add a `log::info!` at the entry of the error check branch in `open_cli` to confirm it's being hit. Verify the binary being tested is the PR build (`plexi-pr-N open ...`), not the stable/beta binary.

ianjamesburke:
## Summary

- Change `launch_app_by_id_with_layout` and `launch_app_by_path_with_layout` return types from `()` to `Result<(), String>`
- `SpawnPane` IPC handler captures the result and writes `{"error": "..."}` to the response file on failure instead of unconditionally writing `{"pane_id": N}`
- `open_cli` and `app_run` check for the `error` field before `pane_id`, exit 1 with `eprintln!` on error
- All non-IPC callers discard the result with `let _ =`

Closes #1542

## Test plan
- [ ] `plexi open <app-with-capability-mismatch>` prints error to stderr and exits 1 — no pane ID printed
- [ ] `plexi open <valid-app>` still prints pane ID and exits 0
- [ ] `plexi app run <path>` follows same contract on error

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1497,6 +1497,7 @@ impl PlexiApp {
                     }
 
                     let cwd_override: Option<std::path::PathBuf> = cwd.as_deref().map(std::path::PathBuf::from);
+                    let mut launch_result: Result<(), String> = Ok(());
                     if type_id == "terminal" {
                         let layout_str = layout.as_deref().unwrap_or("split_v");
                         let initial_cmd = cmd_from_args(args);
@@ -1522,9 +1523,9 @@ impl PlexiApp {
                             self.split_focused(vertical, initial_cmd.as_deref(), *ephemeral, new_pane_first, cwd_override);
                         }
                     } else if let Some(path_str) = path {
-                        self.launch_app_by_path_with_layout(path_str, layout.clone());
+                        launch_result = self.launch_app_by_path_with_layout(path_str, layout.clone());
                     } else {
-                        self.launch_app_by_id_with_layout(type_id, layout.clone(), args, cwd_override);
+                        launch_result = self.launch_app_by_id_with_layout(type_id, layout.clone(), args, cwd_override);
                     }
 
                     // Restore original focus when no_focus is requested or from_pane_id overrode it.
@@ -1538,7 +1539,13 @@ impl PlexiApp {
                         self.windows[active].focused_pane = original_focused;
                     }
                     if let Some(rf) = response_file {
-                        let json = format!("{{\"pane_id\":{new_pane_id}}}");
+                        let json = match &launch_result {
+                            Ok(()) => format!("{{\"pane_id\":{new_pane_id}}}"),
+                            Err(msg) => {
+                                log::warn!("pane_ipc: spawn_pane: launch failed, returning error to caller: {msg}");
+                                format!("{{\"error\":{}}}", serde_json::to_string(msg).unwrap_or_else(|_| format!("\"{msg}\"")))
+                            }
+                        };
                         if let Err(e) = std::fs::write(rf, &json) {
                             log::error!("pane_ipc: spawn_pane: could not write response file: {e}");
                         }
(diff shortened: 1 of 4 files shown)


ianjamesburke:
Closing — CLI exit code check in open_cli doesn't trigger correctly: raw JSON prints to stdout with exit 0 instead of formatted error to stderr with exit 1. The host-side response file format is correct; the cli.rs error field check has a logic bug. Re-opening #1542 as ready for a clean attempt.

PR #1713, opened 2026-05-24
feat(host/pane-ops): send pane in direction — Ctrl+Opt+Cmd+HJKL moves pane without following focus (#1710)

### LINKED ISSUE #1710: feat(host/pane-ops): send pane in direction — Ctrl+Opt+Cmd+HJKL moves pane without following focus
## Summary
Add a 'send pane' shortcut (Ctrl+Opt+Cmd+HJKL) that moves the focused pane in a direction without carrying focus with it — focus stays on the pane that fills the vacated slot.

## What
A complement to the existing swap shortcut (Cmd+Ctrl+HJKL). Both swap two adjacent panes, but they differ in where focus lands:

- **Swap (Cmd+Ctrl+HJKL):** focus follows the pane content (you move with your pane)
- **Send (Ctrl+Opt+Cmd+HJKL):** focus stays at the vacated tile position (your pane moves away; you stay)

Example: you're focused on pane 2 of 5. Press send-right. Pane 2 swaps with pane 3, but focus transfers to pane 3 (now in slot 2). You stay where you were; the pane is gone to the right.

## Why
Useful when reorganizing layout without losing your place. Currently you must swap-then-navigate-back, which is two keystrokes and breaks flow.

## Requirements (updated after attempt 1)
- Must work for **all four directions** (H/J/K/L). J/K vertical failed in attempt 1.
- Must work **both within a tile-split window** (Cmd+D splits) and **across grid windows** (Cmd+N pages). Attempt 1 only worked within a single egui_tiles tree.
- At true boundary (no neighbor in any direction, no adjacent grid window): edge-pulse only.

## Done When
- Pressing Ctrl+Opt+Cmd+L while focused on pane 2 of 3 (tile-split via Cmd+D) moves pane 2 to slot 3; pane from slot 3 is now in slot 2 and is focused.
- Pressing Ctrl+Opt+Cmd+J while focused on top pane of a vertical split sends it down; focus stays at the top slot.
- Pressing Ctrl+Opt+Cmd+L from a grid window (Cmd+N page) sends the pane to the adjacent grid window; focus stays at the source grid window (on whatever pane is now there, or edge-pulses if source becomes empty).
- Swap (Cmd+Ctrl+HJKL) behavior is unchanged.
- `cargo test --bin plexi` green.

## Known Gotchas
- **egui_tiles bare-pane root restructuring:** On first render, a bare `Tile::Pane` root gets converted to a `Tile::Container`. Always run at least one idle frame in tests before reading tile structure or setting `focused_pane`. See `find_pane_in_tile()` usage in existing swap tests.
- **Tile-level vs grid-level:** `find_pane_in_direction_from` only finds neighbors within the same egui_tiles tree (single window). Cross-window send requires `move_focused_pane_to_adjacent_window` plus post-move focus correction: find the pane now occupying the source slot in the source window, focus it.
- **J/K boundary:** `move_focused_pane_to_row_boundary` follows focus by design. For SendPane J/K, either skip the row-boundary move entirely (edge-pulse) or implement a stay-at-source variant.

## Prior Attempts

**Attempt 1:** Implemented `send_pane()` in `layout.rs` as a variant of `swap_pane` with `ctx.focused_pane = Some(focused)` (stay at tile). Bound to Ctrl+Opt+Cmd+HJKL.
**Why it failed:**
- J/K (vertical) did not behave correctly — exact symptom not captured, but tested and failed.
- Cross-grid-window sends (Cmd+N page layout) fell through to `move_focused_pane_to_row_boundary` / `move_focused_pane_to_adjacent_window`, which follow focus with the pane (same as swap).
- At-boundary fix attempted in validate round 2 (edge-pulse instead of move), but J/K and cross-window cases remained unaddressed.
**What to try next:** Handle the cross-window case explicitly: after `AtBoundary`, call `move_focused_pane_to_adjacent_window` and then explicitly refocus the pane now at the source position. For J/K boundary, edge-pulse only (no row-boundary move). For tile-level J/K within a window, verify `find_pane_in_direction_from` correctly returns vertical neighbors — this may require a geometry fix.

### Attempt 1 — 2026-05-24
**Branch:** feature/1710-send-pane-direction
**PR:** #1713 (closed after 2 validate rounds)
**Validate attempt 1:** FAIL — all cases appeared to follow focus; root cause was user layout used Cmd+N grid windows (no tile neighbors), so always hit AtBoundary path which followed focus.
**Validate attempt 2:** FAIL — L/R within tile-split window worked. J/K and cross-grid-window cases did not.
**Hard reject:** 2026-05-24

ianjamesburke:
Closes #1710

## Summary

- Adds `SendPane(Direction)` action bound to `Ctrl+Opt+Cmd+HJKL` — the complement to `SwapPane(Cmd+Ctrl+HJKL)`. Both swap adjacent panes, but `SendPane` keeps focus at the vacated slot (you stay; the pane moves away) while `SwapPane` follows the content
- Wires up the full keybinding stack: `cmd_ctrl_alt()` modifier helper, 4 new `send_pane_*` fields in `KeyBindings` and `KeybindingsConfig`, `apply_override!` entries, conflict-detection array, and input-loop detection (checked before swap block, as it has a more specific modifier set)
- `send_pane()` in `layout.rs` mirrors `swap_pane()` except `ctx.focused_pane = Some(focused)` instead of `Some(neighbor)`; 3 unit tests cover no-focus, at-boundary, and focus-preserved-at-boundary cases

## Done When

- Pressing Ctrl+Opt+Cmd+L while focused on pane 2 of 3 moves pane 2 to slot 3; the pane that was in slot 3 is now in slot 2 and is focused.
- Swap (Cmd+Ctrl+L) behavior is unchanged.
- `cargo test --bin plexi` green.

## Notes

- `egui_tiles::Tiles` has no `set_rect` — tile rects only exist after a render frame, so the Swapped case for `send_pane` can't be exercised in headless unit tests. The focus invariant (`focused_pane = Some(focused)`) is directly visible in the implementation.

## Summary by CodeRabbit

* **New Features**
  * Added pane-send functionality, enabling users to move the focused pane in a specified direction while keeping focus in the original location.
  * Extended keybindings configuration system to support new customizable shortcuts for sending panes in all directions (left, down, up, right).

[Review Change Stack](https://app.coderabbit.ai/change-stack/ianjamesburke/PLEXI/pull/1713?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -3342,6 +3342,44 @@ impl eframe::App for PlexiApp {
                         crate::pane_ops::SwapResult::NoFocus => {}
                     }
                 }
+                Action::SendPane(dir) => {
+                    match self.send_pane(dir) {
+                        crate::pane_ops::SwapResult::Swapped {
+                            rect_a, rect_b, ..
+                        } => {
+                            let now = std::time::Instant::now();
+                            self.pane_anims = vec![
+                                PaneSwapAnim { from: rect_a, to: rect_b, started_at: now },
+                                PaneSwapAnim { from: rect_b, to: rect_a, started_at: now },
+                            ];
+                            self.ctx.request_repaint();
+                        }
+                        crate::pane_ops::SwapResult::AtBoundary => {
+                            let moved = match dir {
+                                crate::keys::Direction::Down => {
+                                    self.move_focused_pane_to_row_boundary(true)
+                                }
+                                crate::keys::Direction::Up => {
+                                    self.move_focused_pane_to_row_boundary(false)
+                                }
+                                _ => self.move_focused_pane_to_adjacent_window(dir),
+                            };
+                            if moved {
+                                self.ctx.request_repaint();
+                            } else if let Some(focused) =
+                                self.windows[self.active_window].focused_pane
+                            {
+                                self.edge_pulse = Some(EdgePulse {
+                                    tile: focused,
+                                    dir,
+                                    started_at: std::time::Instant::now(),
+                                });
+                                self.ctx.request_repaint();
+                            }
+                        }
+                        crate::pane_ops::SwapResult::NoFocus => {}
+                    }
+                }
                 Action::ClosePane => {
                     // If the focused app pane has a non-empty nav stack
                     // (via PushNav), Escape routes NavBack to the app
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -3356,23 +3356,14 @@ impl eframe::App for PlexiApp {
                             ];
                             self.ctx.request_repaint();
                         }
-                        crate::pane_ops::SwapResult::AtBoundary => {
-                            log::info!("action: send_pane at_boundary dir={dir:?}");
-                            let moved = match dir {
-                                crate::keys::Direction::Down => {
-                                    self.move_focused_pane_to_row_boundary(true)
-                                }
-                                crate::keys::Direction::Up => {
-                                    self.move_focused_pane_to_row_boundary(false)
-                                }
-                                _ => self.move_focused_pane_to_adjacent_window(dir),
-                            };
-                            if moved {
-                                self.ctx.request_repaint();
-                            } else if let Some(focused) =
+                        crate::pane_ops::SwapResult::AtBoundary
+                        | crate::pane_ops::SwapResult::NoFocus => {
+                            // Send never moves to a new window — focus can't stay at an empty
+                            // slot. Edge-pulse the boundary instead.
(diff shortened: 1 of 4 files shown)


ianjamesburke:
Closing after 2 failed validate attempts. Full context in issue #1710 body. Feature works for L/R within a tile-split window, but J/K vertical doesn't behave correctly and it doesn't work across grid windows (Cmd+N pages). Issue updated with Prior Attempts.

PR #2065, opened 2026-06-06
refactor(pgap/sdk): explicit protocol constraints — Pinned, Column, Frame padding (#2064)

### LINKED ISSUE #2064: refactor(pgap/sdk): replace ad-hoc layout inference with explicit protocol constraints — Pinned node, Frame padding, Column wire type, schema sync
## Summary
The PGAP layout system has an implicit contract between the Python SDK and host renderer — layout intent is inferred from generic node types rather than declared — making a class of visual bugs structurally inevitable and developer experience fragile.

## Root Cause Analysis

Three confirmed code locations reveal the same underlying problem: **the host renderer makes layout decisions based on heuristics and hardcodes rather than explicit protocol declarations.**

### Bug 1: Left gutter stripe — wrong egui primitive for padding

`src/render/components.rs:97-103`:
```rust
if padding.left > 0.0 {
    ui.indent("stack_left_pad", |ui| {
        events.extend(render_stack(...));
    });
}
```
`ui.indent()` is an egui code-indentation widget that draws a visible left-border gutter. It is not a padding primitive. Any `UiNode::Stack` with `padding.left > 0.0` renders a visible colored stripe on the pane edge. `Column.to_node()` (`sdk/python/plexi_sdk/ui.py:1543`) always sends `padding.left = SPACE_XL` (24.0), so **every `Column`-rooted UI has this bug.**

### Bug 2: Footer not sticky — no positional constraints on the wire

`src/render/components.rs:802-811` (`render_stack`, vertical branch):
```rust
StackDirection::Vertical => {
    ui.vertical(|ui| {
        for (i, child) in children.iter().enumerate() {
            if i > 0 && gap > 0.0 { ui.add_space(gap); }
            events.extend(render_component_tree(ui, child, ...));
        }
    });
}
```
Top-to-bottom loop, no sticky detection. `UiNode::FooterKeys` and `UiNode::Footer` are semantically distinct protocol variants, but they arrive as children of a generic `UiNode::Stack` because `Column.to_node()` (`ui.py:1535`) outputs `"type": "stack"`. The renderer has no sticky-bottom declaration to act on — it can only guess from child node type, which is fragile.

### Bug 3: Double dark bar — visual properties hardcoded in renderer

`src/render/components.rs:532`:
```rust
painter.rect_filled(rect, 0.0, colors.bg_sidebar);
```
`UiNode::FooterKeys` unconditionally fills its entire allocated rect with `colors.bg_sidebar`. When the parent `Column` already provides a background, this produces a double dark bar. The node's own wire definition carries no background field; the renderer invents visual behavior rather than reading it from the protocol.

---

### 1. `Column` loses its semantic identity at the wire boundary

`sdk/python/plexi_sdk/ui.py:1534` — `Column.to_node()` emits `"type": "stack"`. When it crosses the wire the host has no way to know:
- "This is a Column — apply correct padding via Frame, not `ui.indent()`"
- "One of the children is sticky-bottom — pin it to the available rect bottom"
- "The background is already handled — don't paint `bg_sidebar` on `FooterKeys`"

The host invents answers with heuristics. Every new layout primitive will hit the same category of bug until the contract is explicit.

### 2. Dual rendering paths with divergent capabilities

`Column` has two rendering paths:
- `render()` (`ui.py:1497`) — L0 path: Python does full layout math with grow-spacer slack distribution (`ui.py:1507-1512`)
- `to_node()` (`ui.py:1527`) — L1 path: serializes to `UiNode::Stack`, host renders

The L1 path discards grow-spacer information entirely. The host receives a flat stack. Apps that rely on grow spacers for sticky footer positioning break silently on the L1 path.

The North Star (Phase 2): "apps send a tree of semantic nodes; the host handles **ALL** layout." The current system splits layout responsibility between Python (measurement, grow spacers) and the host (egui rendering). These systems cannot agree because Python cannot know egui's available rect.

### 3. Style tokens duplicated, drift-prone

`SPACE_XL = 24.0` lives in both `sdk/python/plexi_sdk/ui.py:58` and `src/ui/style.rs:33`. Confirmed: `src/ui/style.rs:49` has an explicit comment "Keep in sync with RADIUS_BADGE in plexi_sdk/ui.py" — acknowledging drift is already a known risk. Currently maintained by convention, not tooling.

### 4. `UiNode` has a stub `JsonSchema` impl that returns `{}`

`src/protocol/ui_nodes.rs:228-234`:
```rust
impl schemars::JsonSchema for UiNode {
    fn json_schema(_gen: &mut schemars::SchemaGenerator) -> schemars::Schema {
        schemars::json_schema!({})  // ← empty schema
    }
}
```
Added to avoid schemars recursion issues with `Box<UiNode>`. As a result, no machine-readable contract exists for the PGAP wire format. The Python SDK cannot be validated against the protocol, and any new node type added to Rust silently never appears in generated documentation or tooling.

---

### Track A — `UiNode::Pinned`: sticky-bottom as an explicit protocol constraint

Add `UiNode::Pinned { edge: PinnedEdge, child: Box<UiNode> }` to the protocol. `FooterKeys` and `Footer` in the Python SDK emit themselves wrapped in `Pinned { edge: Bottom }`. The host renderer extracts `Pinned { edge: Bottom }` children from a vertical stack before rendering the body, then renders them anchored to `ui.available_rect_before_wrap().max.y`.

**Why `Pinned` over node-type inference:** inference requires updating `render_stack()` every time a new sticky-bottom node type is added. `Pinned` is a general mechanism — any node type can be pinned with zero renderer changes. This makes the class of "node doesn't pin to bottom" bugs impossible by construction.

### Track B — `egui::Frame` margin: correct padding primitive (smallest / fastest)

Replace the `ui.indent()` block with `egui::Frame::new().inner_margin()` for all Stack padding. This is a ~15-line change in one file with zero protocol changes. The left stripe disappears for every consumer immediately.

```rust
// Before (components.rs:92-108)
ui.scope(|ui| {
    if padding.top > 0.0 { ui.add_space(padding.top); }
    if padding.left > 0.0 {
        ui.indent("stack_left_pad", |ui| { render_stack(...) });
    } else {
        render_stack(...)
    }
    if padding.bottom > 0.0 { ui.add_space(padding.bottom); }
});

// After
egui::Frame::new()
    .inner_margin(egui::Margin {
        left: padding.left,
        right: padding.right,
        top: padding.top,
        bottom: padding.bottom,
    })
    .show(ui, |ui| {
        events.extend(render_stack(ui, direction, children, *gap, colors, text_edit_buffers, focus_ctx));
    });
```

### Track C — `UiNode::Column` as a distinct wire type

Add `UiNode::Column { children: Vec<UiNode>, gap: f32, padding_top: f32 }` as a new protocol variant. Left/right/bottom padding are implied constants (`SPACE_XL`) — the type carries its semantic contract. `Column.to_node()` emits `"type": "column"` instead of `"type": "stack"`, dropping the explicit `padding` dict.

**Why:** `Column` carries semantic identity (standard app margins, sticky-footer host contract). Collapsing it to `UiNode::Stack` destroys this identity, forcing the renderer to re-derive it from hardcodes. A distinct protocol type means the renderer handles `Column` correctly by definition. Also enables future column-specific features (column-level scroll, column-level theming) without protocol changes.

**Interaction with Track A:** with `Pinned` in place, `Column`'s renderer arm just needs to apply Frame margin and render children in normal flow order — sticky behavior is handled by `Pinned` nodes within the children, not by Column-specific logic.

### Track D — Real `UiNode` schema + style token codegen (phased, larger)

**D1 — Fix the stub `JsonSchema` impl (`src/protocol/ui_nodes.rs:228-234`):**
The `Box<UiNode>` recursion issue is solvable. Options: (a) use `schemars`'s `SchemaGenerator` context which handles recursive refs via `$defs` automatically; (b) use `#[schemars(schema_with = "...")]` on recursive fields. Confirm the correct approach with a standalone test before wiring into build.

**D2 — Export schema at build time (`build.rs`):**
Add a step that serializes `schemars::schema_for!(UiNode)` to `sdk/pgap_schema.json` on every build. The schema becomes the machine-readable protocol contract — used for documentation, validation, and Python type generation.

**D3 — Generate Python style tokens from Rust (`src/ui/style.rs`):**
Add a build step (or standalone binary) that exports `SPACE_*`, `TEXT_*`, `RADIUS_*` constants from `src/ui/style.rs` to `sdk/python/plexi_sdk/_style_tokens.py`. Replace the manual `SPACE_XL = 24.0` etc. in `ui.py:50-76` with imports from `_style_tokens`. Drift between Python and Rust style values becomes a build error.

**D4 — Generate Python SDK base types from schema (optional, future):**
Use `datamodel-codegen` to generate typed Python dataclasses from `pgap_schema.json`. The hand-rolled `to_node()` methods become thin wrappers over generated types. Adding a new `UiNode` variant in Rust automatically produces a Python type on next codegen run. Protocol drift becomes a CI failure.

---

## Implementation Map

```
src/protocol/ui_nodes.rs
  ~line 32  — add PinnedEdge enum { Bottom, Top, Left, Right }
  ~line 64  — add UiNode::Pinned { edge: PinnedEdge, child: Box<UiNode> } variant
  ~line 64  — add UiNode::Column { children: Vec<UiNode>, gap: f32, padding_top: f32 } variant
  lines 228-234 — fix stub JsonSchema impl (Track D1)
  lines 243-308 — add Pinned and Column arms to manual PartialEq impl

src/render/components.rs
  lines 92-108   — replace ui.indent() with egui::Frame::new().inner_margin() (Track B)
  lines 781-814  — add Pinned extraction pass before vertical render_stack loop (Track A)
  line 532       — remove unconditional bg_sidebar fill from FooterKeys renderer
  ~line 87       — add UiNode::Pinned match arm: extract and defer to bottom pass
  ~line 87       — add UiNode::Column match arm: apply SPACE_XL Frame margin, call render_stack

sdk/python/plexi_sdk/ui.py
  line 1534      — Column.to_node(): change "type": "stack" → "type": "column", remove padding dict, keep gap + padding_top (Track C)
  line 1543      — remove explicit "left": self.padding from padding dict (no longer needed after Track C)
  ~line 826      — FooterKeys.to_node(): wrap returned dict in {"type": "pinned", "edge": "bottom", "child": <inner>} (Track A)
  ~line 743      — Footer.to_node(): same Pinned wrapper (Track A)
  lines 58-76    — replace hardcoded SPACE_* / TEXT_* / RADIUS_* with imports from _style_tokens (Track D3)

build.rs
  — add pgap_schema.json export step (Track D2)
  — add style_tokens generation step (Track D3)
```

**Do not touch:** `src/process_app/render_session.rs` (wires ComponentTree to render_component_tree, no layout logic), `sdk/python/plexi_sdk/_render_context.py` (calls render_tree(), no layout changes needed), any app code in `apps/` (layout changes must be invisible to app authors — verify by running the counter test app after each track).

---

## Done When

1. Any app using `Column([..., FooterKeys([...])])` with no code changes renders: zero left gutter stripe, zero double dark bar, `FooterKeys` pinned flush to pane bottom.
2. A `HostHarness` test `render_footer_keys_pins_to_bottom` passes: short content + `FooterKeys` → footer rect bottom equals available pane rect bottom.
3. Adding a new sticky-bottom node in the future requires only: emit `Pinned { edge: Bottom }` in the node's `to_node()` — zero changes to `render_stack()`, zero changes to `components.rs`.
4. `Column` padding does not call `ui.indent()` anywhere — `egui::Frame` handles all margin.
5. `cargo test --bin plexi` passes.
6. (Track D) `sdk/pgap_schema.json` exists and contains a complete, non-empty schema for `UiNode`.

## Open Questions

- **Track D1 approach:** confirm whether `schemars`'s `SchemaGenerator::subschema_for::<UiNode>()` resolves the `Box<UiNode>` recursion via `$defs` automatically in schemars 1.x, or whether the manual impl needs to hand-craft the `$defs` block. Write a standalone test before wiring into build.rs.
- **Track C necessity after Track A:** with `Pinned` handling sticky behavior, does `UiNode::Column` still need to be a distinct type, or is `UiNode::Stack` + `egui::Frame` padding sufficient? Recommend keeping `Column` as a distinct type for semantic explicitness, but this should be confirmed before implementing Track C.

---

_Supersedes #2063 (three visual bugs are symptoms of the root causes addressed here). Tracks B and A are prerequisites for everything else; Track C and Track D can be parallelized after Track A lands._

### Attempt 1 — 2026-06-05
**Branch:** feature/2064-pinned-column-frame-padding
**Files changed:** src/protocol/ui_nodes.rs, src/render/components.rs, sdk/python/plexi_sdk/ui.py
**Spec summary:** Track A (Pinned protocol + bottom-hoist in render_stack), Track B (Frame inner_margin replaces ui.indent), Track C (Column wire type), Bug3 (remove FooterKeys bg_sidebar fill), Track D1 (UiNode derives JsonSchema)
**PR:** #2065 — https://github.com/ianjamesburke/PLEXI/pull/2065
**AI Review:** 0 fixes | no changes needed (CodeRabbit local: clean, bot: auto-skipped)
**Validate attempt 1:** PASS
**Test date:** 2026-06-05
**Merged:** PR #2065 → alpha v0.0.644 (2026-06-05)

ianjamesburke:
Closes #2064

## Summary

- **Track A:** Add `UiNode::Pinned{edge,child}` — vertical stacks hoist `Pinned{Bottom}` children, constraining body height so footers render flush to the available rect bottom
- **Track B:** Replace `ui.indent()` with `egui::Frame::inner_margin()` for Stack padding — eliminates the left gutter stripe on every `Column`-rooted UI
- **Track C:** Add `UiNode::Column{children,gap,padding_top}` as a distinct wire type; `Column.to_node()` now emits `"type":"column"` instead of collapsing to a generic stack
- **Bug 3:** Remove unconditional `bg_sidebar` fill from `FooterKeys` renderer (was causing double dark bar when Column provided background)
- **Track D1:** `UiNode` now derives `JsonSchema` (was returning empty `{}` stub); schemars 1.x handles `Box<UiNode>` recursion via `$defs`
- **Python SDK:** `Footer.to_node()` and `FooterKeys.to_node()` wrap in `Pinned{edge:bottom}`; `Column.to_node()` emits column type, drops explicit padding dict

## Done When

1. Any app using `Column([..., FooterKeys([...])])` renders: zero left gutter stripe, zero double dark bar, `FooterKeys` pinned flush to pane bottom.
2. A `HostHarness` test `render_footer_keys_pins_to_bottom` passes: short content + `FooterKeys` → footer rect bottom equals available pane rect bottom.
3. Adding a new sticky-bottom node requires only: emit `Pinned { edge: Bottom }` in the node's `to_node()` — zero changes to `render_stack()`.
4. `Column` padding does not call `ui.indent()` anywhere — `egui::Frame` handles all margin.
5. `cargo test --bin plexi` passes.
6. `sdk/pgap_schema.json` exists with a complete non-empty schema for `UiNode`.

## Notes

- `bottom_pin_height()` helper computes fixed heights for `FooterKeys` and `Footer` (the two known bottom-pinnable types); unknown nodes return `None` and render inline
- Track D2 (schema export at build time) and Track D3 (style token codegen) are deferred — Track D1 (fix JsonSchema stub) is included
- The one pre-existing failing test (`render_events_coalesced_non_render_events_preserved`) was already failing on `origin/alpha` before this branch

### DIFF
--- a/sdk/python/plexi_sdk/ui.py
+++ b/sdk/python/plexi_sdk/ui.py
@@ -741,8 +741,8 @@ class Footer(Component):
                      size=TEXT_HINT, color=self.color or theme.muted)
 
     def to_node(self) -> dict:
-        return {"type": "footer", "text": self.text,
-                "color": self.color or ""}
+        inner = {"type": "footer", "text": self.text, "color": self.color or ""}
+        return {"type": "pinned", "edge": "bottom", "child": inner}
 
 
 @dataclass
@@ -831,7 +831,8 @@ class FooterKeys(Component):
             else:
                 keys = list(keys_or_key)
             entries.append({"keys": keys, "description": desc})
-        return {"type": "footer_keys", "entries": entries, "divider": self.divider}
+        inner = {"type": "footer_keys", "entries": entries, "divider": self.divider}
+        return {"type": "pinned", "edge": "bottom", "child": inner}
 
 
 # ── Composite list components ──────────────────────────────────────────────
@@ -1532,16 +1533,10 @@ class Column(Component):
                 return None
             children.append(node)
         return {
-            "type": "stack",
-            "direction": "vertical",
+            "type": "column",
             "children": children,
             "gap": self.gap,
-            "padding": {
-                "top": self._pad_top,
-                "right": self.padding,
-                "bottom": self.padding,
-                "left": self.padding,
-            },
+            "padding_top": self._pad_top,
         }
 
 
--- a/sdk/python/plexi_sdk/ui.py
+++ b/sdk/python/plexi_sdk/ui.py
@@ -1460,13 +1460,9 @@ class Column(Component):
     any `Spacer(grow=True)` descendants at the top level.
 
     Padding defaults to `SPACE_XL` (24px) on the sides and bottom, and
-    `SPACE_SM` (8px) on the top. A top-of-pane `Header` carries its own
-    visual weight via TEXT_TITLE_XL and its own bottom rhythm (gap +
-    divider), so anything above a few px reads as "the title is dropped"
-    rather than anchored. The other three sides stay at 24px where content
-    *does* need breathing room.
-
-    Override either with `padding=` (all sides) or `padding_top=` (top only).
+    `SPACE_SM` (8px) on the top. Pass `padding=0` for full-width content
+    (e.g. apps whose children manage their own horizontal margins).
+    Override top-only with `padding_top=`.
     """
     children: List[Component]
     padding: float = SPACE_XL
@@ -1537,6 +1533,7 @@ class Column(Component):
             "children": children,
             "gap": self.gap,
             "padding_top": self._pad_top,
+            "padding": self.padding,
         }
 
 
--- a/src/protocol/ui_nodes.rs
+++ b/src/protocol/ui_nodes.rs
@@ -8,6 +8,10 @@ fn default_true() -> bool {
     true
 }
 
+fn default_space_xl() -> f32 {
+    24.0 // SPACE_XL — keep in sync with src/ui/style.rs
+}
+
 /// Single shortcut entry for `UiNode::FooterKeys`.
 #[derive(Serialize, Deserialize, JsonSchema, Debug, Clone, PartialEq)]
 pub struct FooterKeyEntry {
@@ -120,14 +124,17 @@ pub enum UiNode {
         edge: PinnedEdge,
         child: Box<UiNode>,
     },
-    /// Semantic column container — standard app margins (SPACE_XL left/right),
-    /// sticky-footer host contract. Emitted by `Column.to_node()` in the Python SDK.
+    /// Semantic column container with sticky-footer host contract.
+    /// `padding` sets left/right/bottom margins (default SPACE_XL); `padding_top`
+    /// sets the top margin independently. Emitted by `Column.to_node()` in the SDK.
     Column {
         children: Vec<UiNode>,
         #[serde(default)]
         gap: f32,
         #[serde(default)]
         padding_top: f32,
+        #[serde(default = "default_space_xl")]
+        padding: f32,
     },
 
     // ── L1 sugar ─────────────────────────────────────────────────────────
@@ -317,9 +324,9 @@ impl PartialEq for UiNode {
             (UiNode::Pinned { edge: e1, child: c1 }, UiNode::Pinned { edge: e2, child: c2 }) => {
                 e1 == e2 && c1 == c2
             }
-            (UiNode::Column { children: c1, gap: g1, padding_top: p1 },
-             UiNode::Column { children: c2, gap: g2, padding_top: p2 }) => {
-                c1 == c2 && g1 == g2 && p1 == p2
+            (UiNode::Column { children: c1, gap: g1, padding_top: p1, padding: pa1 },
+             UiNode::Column { children: c2, gap: g2, padding_top: p2, padding: pa2 }) => {
+                c1 == c2 && g1 == g2 && p1 == p2 && pa1 == pa2
             }
             _ => false,
         }
--- a/src/protocol/ui_nodes.rs
+++ b/src/protocol/ui_nodes.rs
@@ -54,12 +54,19 @@ pub struct UiPadding {
     pub left: f32,
 }
 
+/// Which edge to pin a child against within a vertical Stack.
+#[derive(Serialize, Deserialize, JsonSchema, Debug, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum PinnedEdge {
+    Bottom,
+    Top,
+    Left,
+    Right,
+}
+
 /// Component tree node. L0 primitives compose into rich UI; L1 sugar variants
 /// are rendered natively by the host.
-///
-/// `JsonSchema` is implemented manually to avoid schemars recursion issues
-/// (UiNode contains Box<UiNode>).
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, JsonSchema, Debug, Clone)]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum UiNode {
     // ── L0 primitives ────────────────────────────────────────────────────
@@ -107,6 +114,21 @@ pub enum UiNode {
     Raw { command: Box<RenderCommand> },
     /// Future GPU surface placeholder — reserved, not yet rendered.
     Surface { id: String },
+    /// Pinned layout wrapper. In a vertical Stack, `Pinned { edge: Bottom }` children
+    /// are rendered flush to the available rect bottom regardless of body content height.
+    Pinned {
+        edge: PinnedEdge,
+        child: Box<UiNode>,
+    },
+    /// Semantic column container — standard app margins (SPACE_XL left/right),
+    /// sticky-footer host contract. Emitted by `Column.to_node()` in the Python SDK.
+    Column {
+        children: Vec<UiNode>,
+        #[serde(default)]
+        gap: f32,
+        #[serde(default)]
+        padding_top: f32,
+    },
 
     // ── L1 sugar ─────────────────────────────────────────────────────────
     /// Host-rendered button.
@@ -224,16 +246,6 @@ pub enum UiNode {
     },
 }
 
-// Manual JsonSchema impl to avoid schemars recursion (UiNode contains Box<UiNode>).
-impl schemars::JsonSchema for UiNode {
-    fn schema_name() -> std::borrow::Cow<'static, str> {
-        "UiNode".into()
-    }
-    fn json_schema(_gen: &mut schemars::SchemaGenerator) -> schemars::Schema {
-        schemars::json_schema!({})
-    }
-}
-
 // Manual PartialEq impl — UiNode::Raw wraps RenderCommand which would require
 // PartialEq on hundreds of transitive types. Raw nodes are structural escape
 // hatches; we compare them via their serde round-trip JSON representation so
@@ -302,6 +314,13 @@ impl PartialEq for UiNode {
              UiNode::Card { children: c2, padding: p2 }) => c1 == c2 && p1 == p2,
             (UiNode::SelectList { items: i1, selected_idx: s1 },
              UiNode::SelectList { items: i2, selected_idx: s2 }) => i1 == i2 && s1 == s2,
+            (UiNode::Pinned { edge: e1, child: c1 }, UiNode::Pinned { edge: e2, child: c2 }) => {
+                e1 == e2 && c1 == c2
+            }
+            (UiNode::Column { children: c1, gap: g1, padding_top: p1 },
+             UiNode::Column { children: c2, gap: g2, padding_top: p2 }) => {
+                c1 == c2 && g1 == g2 && p1 == p2
+            }
             _ => false,
         }
     }
--- a/src/render/components.rs
+++ b/src/render/components.rs
@@ -7,7 +7,7 @@
 
 use egui::Ui;
 
-use crate::app_protocol::{StackDirection, UiNode};
+use crate::app_protocol::{PinnedEdge, StackDirection, UiNode};
 use crate::ui::style;
 use crate::ui::theme::Colors;
 
@@ -90,21 +90,16 @@ pub(crate) fn render_component_tree(
         // ── L0 primitives ────────────────────────────────────────────────
 
         UiNode::Stack { direction, children, gap, padding } => {
-            ui.scope(|ui| {
-                if padding.top > 0.0 {
-                    ui.add_space(padding.top);
-                }
-                if padding.left > 0.0 {
-                    ui.indent("stack_left_pad", |ui| {
-                        events.extend(render_stack(ui, direction, children, *gap, colors, text_edit_buffers, focus_ctx));
-                    });
-                } else {
+            egui::Frame::new()
+                .inner_margin(egui::Margin {
+                    left: padding.left as i8,
+                    right: padding.right as i8,
+                    top: padding.top as i8,
+                    bottom: padding.bottom as i8,
+                })
+                .show(ui, |ui| {
                     events.extend(render_stack(ui, direction, children, *gap, colors, text_edit_buffers, focus_ctx));
-                }
-                if padding.bottom > 0.0 {
-                    ui.add_space(padding.bottom);
-                }
-            });
+                });
         }
 
         UiNode::Scroll { child, horizontal } => {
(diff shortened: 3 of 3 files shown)


ianjamesburke:
A few concerns from a closer look — flagging the ones that look load-bearing to me rather than nits:

**1. `bottom_pin_height` is a render-contract leak into layout code, and it will silently break as soon as styles change.**

```rust
UiNode::FooterKeys { divider, .. } => {
    let chip_row_h = style::TEXT_HINT + 6.0;
    Some(if *divider {
        style::SPACE_SM + 1.0 + style::SPACE_SM + chip_row_h + style::SPACE_SM
    } else {
        style::SPACE_SM + chip_row_h + style::SPACE_SM
    })
}
```

This duplicates the geometry math that already lives in the `FooterKeys` renderer (rect allocation, divider offset, chip row, padding). If anyone tweaks the footer chrome — say, adding an extra `SPACE_XS` margin, or the `+6.0` chip padding ever becomes a `+8.0` — this constant drifts silently and the footer will float up by exactly that delta. The renderer will still look fine, but `body_h` will be wrong and the "pin to bottom" invariant breaks invisibly.

Stronger fix: render the pinned child first into a `Ui::child_ui` with `Sense::hover()` (or measure via `galley`/`text_layout` for the text-only case), read `desired_size.y`, then allocate the body with `available_height() - that`. Yes, it is a real render pass, but the height is in one place. Alternatively, hoist the height math into an associated function on a `Pinnable` trait that both the renderer and the layout code call, so the geometry has a single source of truth.

**2. `render_stack` now has two divergent "vertical" code paths in one match arm.**

```rust
if !pinned_bottom.is_empty() {
    // body + pinned
} else {
    ui.vertical(|ui| { ... });
}
```

The two branches are doing almost the same thing — iter children, gap, recurse. The pinned branch additionally partitions. Fine. But the partition logic is interleaved with render, and the `if let ... if let Some(h)` chain quietly drops pins for any node whose `bottom_pin_height` returns `None`. So if I emit `Pinned { edge: Bottom, child: Label }` from Python, my pin is silently ignored and the label renders inline. That is a footgun: the protocol says "pin" but the host says "no I won't" without any error.

I would at minimum `log::warn!` when a `Pinned` node is skipped due to unknown height, so this surfaces in `plexi.log` instead of being a silent contract violation. Even better: have `Pinned` carry an explicit `intrinsic_size: Option<f32>` from the SDK so the protocol is honest about when the height is known.

**3. The `Stack` padding rewrite loses `ui.indent()` but the `ui.indent` is *load-bearing* in the old code for a reason — worth verifying, not just trusting the test.**

Old:
```rust
ui.indent("stack_left_pad", |ui| { render_stack(...); });
```

New:
```rust
egui::Frame::new().inner_margin(...).show(ui, |ui| { render_stack(...); });
```

`ui.indent` does not just push the content right — it narrows the available width for child layout. `Frame::inner_margin` does too, so it should be equivalent for monospace text and labels. But: any child that reads `ui.cursor().left` or computes positioning relative to the *unindented* origin will see a different value. The new code also moves the inner content to a fresh `Ui` with a clip rect that excludes the margin. The "zero left gutter stripe" goal is real, but the test that proves it (`render_footer_keys_pins_to_bottom`) only checks the footer rect — it does not check that `ui.indent()` callers downstream still work. If `Stack { padding.left: 24 }` is used anywhere with a child that does its own positioning, this is a silent visual change.

I would want a grep for `Stack {` with non-zero `padding` in Python SDK usage and a screenshot pass against at least one app that uses `Stack` with padding (not `Column`) before merging. The PR only documents `Column` behavior, but the `Stack` rewrite is a separate change with separate test coverage.

**4. The `Column.to_node()` change drops the `padding` dict entirely — that is a wire-format break for any non-SDK caller.**

Old wire:
```json
{"type": "stack", "direction": "vertical", "padding": {"top": 8, "right": 24, "bottom": 24, "left": 24}, ...}
```

New wire:
```json
{"type": "column", "padding_top": 8, ...}
```

The host's `UiNode::Column` only honors `SPACE_XL` for left/right and `0` for bottom — the Python `Column.padding` default of `SPACE_XL` is *implicit* in the renderer now, not in the wire. If anyone (a third-party app, a recorded replay, a test fixture) constructs the old wire format by hand, it will deserialize as `UiNode::Column` with the `padding` field gone, and the `left`/`right` 24px margin silently appears where it might not have been.

Two options: (a) document the wire break loudly in the PR description and in `CHANGELOG.md`; (b) keep the old `padding` field as `#[serde(default, skip_serializing_if = "Option::is_none")]` and warn-on-ignore on the host side. Given there are no non-SDK producers today, (a) is probably fine — but the PR's "wire format" framing implies an upgrade path that I do not see called out.

**5. `bg_sidebar` removal in `FooterKeys` is correct, but only if the parent always provides background. Worth checking.**

```rust
// Before:
painter.rect_filled(rect, 0.0, colors.bg_sidebar);
// After: (removed)
```

The PR's bug 3 framing — "remove unconditional fill to avoid double dark bar" — assumes that *somebody upstream* draws a background. `Column` does (`Frame::inner_margin` with default frame has no fill, but `Frame` defaults to `fill: Default::default()` which is transparent — so `Column` does NOT draw a background).

This means a bare `Stack` containing `FooterKeys` at the root will now render with no background at all. Previously it had `bg_sidebar` dark fill. If any app uses `Stack` (not `Column`) with a `FooterKeys` child, the footer's background just disappears. That is a *visual* regression, not a fix. The test only covers the `Column` path.

I would either keep a fallback `bg_sidebar` fill in the renderer when no ancestor provided a background, or document that `FooterKeys` requires a `Column` (or other background-providing) parent. The latter is cleaner and aligns with the "explicit protocol constraints" framing, but it is a contract change that should be in the PR description.

**6. Minor: `padding_top` and `gap` are `#[serde(default)]` on `Column`, but `Column.to_node()` always sends them. Inconsistency is fine for forward compat, but `JsonSchema` will now show them as optional in the schema — worth verifying the schema is what you want agents to see.**

The Track D1 schema work is good. The deferred Track D2 (build-time export) and D3 (style codegen) are the right calls to defer. But: with `JsonSchema` now deriving correctly, the schema for `UiNode::Pinned` will be a `$ref` to a definition, and the schema for `PinnedEdge` will be inline. If you have any other tool that consumes the schema (e.g. a TS codegen step in another repo), verify it can handle the `$defs`/`$ref` indirection. The PR description says "schemars 1.x handles it" but the consumer is the contract.

**7. `PartialEq` manual impl is now bigger — is the new arm order intentional?**

```rust
(UiNode::SelectList { ... },
 UiNode::SelectList { ... }) => ...,
(UiNode::Pinned { ... }, UiNode::Pinned { ... }) => ...,
(UiNode::Column { ... }, UiNode::Column { ... }) => ...,
_ => false,
```

Pattern match order does not matter for correctness here (each arm is a distinct variant), but adding new variants to `UiNode` requires touching two places (the enum + this `match`) and the `#[deny(clippy::todo)]` lint will catch missing arms at compile time. Fine. Just calling out that the manual `PartialEq` is now ~30 lines of variant-by-variant arms and is starting to be a maintenance liability. Long-term: derive `PartialEq` on a `RenderCommand`-less view-type, or compare via `serde_json::Value` (which you already do for `Raw`). The current "structural escape hatch for Raw" comment understates how much manual code this is accumulating.

---

Overall the intent and the test design are sound — the bottom-pin invariant is the right thing to lock in. The concerns above are mostly about ensuring the test suite covers the *non-happy* paths (Stack with padding, `Pinned` with unknown height, `FooterKeys` in a non-`Column` parent) before the wire format becomes load-bearing for replays and external producers.

PR #2296, opened 2026-06-19
feat(wasm): zero-copy surface present + host frame clock/telemetry

ianjamesburke:
## Summary
Makes WASM GPU surfaces a high-performance app surface by removing the per-frame readback from the live present path.

- **Shared device**: register eframe's `RenderState` at startup; WASM `GpuDevice::from_shared` so the guest surface lives on egui's device. Headless/tests fall back to a dedicated device.
- **Zero-copy present**: register the surface as an `Rgba8UnormSrgb` view into egui's renderer once (`register_native_texture`) and sample it live. Per-frame readback + upload + both hot-path `info!` logs removed. Readback survives only as the no-shared-device fallback.
- **WasmFrameClock**: host-owned fixed timestep with bounded catch-up + drop policy.
- **FrameTelemetry**: host-owned fps / p95-interval / p95-present / dropped, sampled every 120 frames. Apps display host telemetry; they never invent FPS.
- **Perf gate**: deterministic per-device readback counter — 300 steady ticks read back zero; capture path stays isolated (one call, one readback). Plus FrameClock/Telemetry unit tests.

## Design notes
Dropped the speculative 3-adapter compositor for a 2-path split (present vs capture); readback becomes the fallback, which resolved every dead-code warning honestly instead of with `allow`.

## Verification
- `cargo test --bin plexi`: 1283 passed / 0 failed (g7 surface lifecycle + paddle pixel tests intact via fallback).
- `cargo build` + clippy clean on all changed files.
- **Needs human check**: live zero-copy display on a GPU app (bevy-pong) for steady 60fps + correct colors — the deterministic gate proves no-readback/registration but not the felt visual result.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -717,6 +717,16 @@ impl PlexiApp {
             *slot = Some(cc.egui_ctx.clone());
         }
 
+        // Register eframe's shared wgpu device so live WASM surfaces composite
+        // zero-copy (no per-frame readback). Absent only if eframe was built
+        // without the wgpu backend, in which case WASM GPU apps fall back to a
+        // dedicated device + readback path.
+        if let Some(render_state) = cc.wgpu_render_state.clone() {
+            crate::host::wasm_gpu::register_host_render_state(render_state);
+        } else {
+            log::warn!("no wgpu render state at startup; WASM surfaces use readback fallback");
+        }
+
         #[cfg(target_os = "macos")]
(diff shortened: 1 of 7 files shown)


ianjamesburke:
Superseded by #2297, which merges this branch in full (zero-copy present) plus the pong/breakout rename and WASM launch-args work. Closing in favor of the consolidated PR.

PR #2316, opened 2026-06-21
fix: todo app space-to-toggle regression (stint 0281)

ianjamesburke:
## Summary

- Space key in todo (and snake) apps was double-firing: egui emits both `Event::Key { Space }` and `Event::Text(" ")` for a single press
- The prior fix (#2280) removed `Space` from `is_printable_key` so the Key path fires — but the Text path (`' '`) was never suppressed, causing two `on_key("space")` calls per press
- Double-toggle meant items toggled on then immediately off, appearing as "space does nothing"
- Fix: suppress `' '` in the `Event::Text` handler since space is already delivered canonically via the Key path as `"Space"` → SDK alias → `"space"`

## Done When

- Space toggles todo items (checked ↔ unchecked) with a single press
- No double-toggle on any item state

### DIFF
--- a/src/process_app/mod.rs
+++ b/src/process_app/mod.rs
@@ -2397,6 +2397,11 @@ impl App for ProcessApp {
                         if ch.is_control() {
                             continue;
                         }
+                        // Space comes through Event::Key as "Space" (→ SDK alias "space");
+                        // suppress the duplicate Text path to prevent double on_key calls.
+                        if ch == ' ' {
+                            continue;
+                        }
                         // Suppress j/k text events when a ListView is active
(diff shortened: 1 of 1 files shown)


ianjamesburke:
Superseded. This patches `src/process_app/mod.rs`, the PGAP subprocess runtime, which no longer exists in alpha — it was deleted wholesale by the WASM migration (stint 0285, completed 2026-07-12). The diff cannot apply to any current file. Stint 0281 is archived.

If space-to-toggle regresses under the WASM runtime it is a fresh bug against a different code path, and gets a new stint.

PR #2318, opened 2026-06-21
fix(stats): filter idle heartbeats from switches tile and waveform

ianjamesburke:
Stint: 0282

## Summary

- The stats page showed phantom activity overnight because `wave_count[b]` and `m.switches_today` incremented for every event in the day window — including 15-min timer heartbeats.
- `_counted_duration` already zeroes idle-filtered heartbeats (returns `0` after the first clamped entry), but the compute loop ignored those zeros for the waveform and switch counter.
- Fix: gate `wave_count[b]` and `wave_ctx` on `secs > 0`; only count `reason == "pane_switch"` events toward `m.switches_today`. Removes `max(secs, 1.0)` which was painting the waveform even for 0-second events.
- Added two regression tests in `apps/stats/tests/test_phantom_activity.py`.

## Done When

- Overnight idle (no keypresses, no focus switches) produces zero waveform spokes and zero Switches count
- `switches_today` reflects only user-initiated pane switches

### DIFF
--- a/apps/stats/stats.py
+++ b/apps/stats/stats.py
@@ -249,13 +249,18 @@ class Metrics:
             if local < day_start or local >= day_end:
                 continue
             # Within the selected 4 AM-anchored window.
-            m.switches_today += 1
+            # Only user-initiated pane switches count toward the Switches tile;
+            # timer heartbeats are not user activity.
+            if (ev.get("reason") or "") == "pane_switch":
+                m.switches_today += 1
             label = _context_label(ev)
             offset = (local - day_start).total_seconds()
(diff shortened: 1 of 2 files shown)


ianjamesburke:
Superseded. Idle-heartbeat filtering already landed in alpha by another route: `apps/stats/stats.py` carries `IDLE_THRESHOLD_SECS`/`IDLE_CLAMP_SECS` with skip/clamp states, and `apps/stats/tests/test_idle_normalization.py` covers it. This branch is 467 commits behind and would only re-litigate a solved problem. Stint 0282 is archived.

PR #2363, opened 2026-07-06
fix: host start windowless boot drops seeded/spawned panes (stint 0348)

ianjamesburke:
Stint task 0348 (no linked GitHub issue). Found live-driving the event-bus PR validation.

## Summary

- On a windowless `plexi host start` boot, the spawn_pane IPC fallback called `split_focused`, which returns early with no focused pane — seeded (`--pane`) and socket (`pane new`) spawns silently vanished ("could not apply name to pane_id=1", `pane list` → `[]`).
- New `seed_root_pane` populates the empty active window with a root terminal in place, returns the pane id actually created; the response file and inline naming use the real id; seeding failure propagates as an error instead of a phantom success.

## Test evidence (attempt 1)

- cargo test: 1504 passed, 0 failed — full bin suite (×2 consecutive runs). New regression test `spawn_pane_seeds_root_in_empty_window` (failed pre-fix).
- Conclusion: binary install required — the repro is the installed `host start` boot path; validating via `just pr-install` + windowless-boot drive (hands-off).

### DIFF
--- a/src/app/lifecycle.rs
+++ b/src/app/lifecycle.rs
@@ -948,29 +948,43 @@ impl PlexiApp {
                                     {
                                         (active, tile)
                                     } else {
-                                        // Window is empty — let split_focused handle it
-                                        log::info!("pane_ipc: spawn_pane terminal layout={layout_str} (empty context fallback)");
-                                        self.split_focused(
-                                            vertical,
+                                        // Active window is empty (windowless boot):
+                                        // split_focused would no-op, so seed a root
+                                        // pane in place instead of dropping the spawn.
+                                        log::info!("pane_ipc: spawn_pane terminal layout={layout_str} (empty active window — seeding root)");
+                                        match self.seed_root_pane(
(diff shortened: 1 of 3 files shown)


ianjamesburke:
## Live validation — hands-off windowless-boot drive (plexi-pr-2363)

| Check | Result |
|---|---|
| Windowless boot + socket `pane new` | ✅ pane created (was silently dropped before); response returned the real id; inline name applied |
| `host start --pane 'cwd=/tmp' --pane 'cwd=/tmp,cmd=…'` seeding | ✅ both seeded panes materialized with requested cwd; seeded cmd ran |
| Full suite | ✅ 1504 passed ×2 consecutive runs; regression test `spawn_pane_seeds_root_in_empty_window` failed pre-fix |

Minor cosmetic note (not a blocker): `host start` readiness reports `0 pane(s) ready` because the spawn-queue drains on the first frame after the readiness poll ends; seeds land ~1s later.

PR #2405, opened 2026-07-16
fix(apps): port sudoku to current SDK — canvas size + hit_region drift

ianjamesburke:
No linked GitHub issue — tracked as stint task 0394.

## Summary

- Sudoku crashed unconditionally on its first `view()` (`AttributeError: module 'plexi_sdk' has no attribute 'canvas_width'`). `sdk.canvas_width`/`canvas_height`/`pane_width`/`pane_height` were removed as SDK globals; `CanvasRect` no longer accepts `border_color`/`border_width`/`hit_region` kwargs.
- Ported to breakout's proven pattern: learn canvas size via `init(size)` + `Resize`, and pass `width=`/`height=` to `Canvas()` so the declared coordinate space matches the actual render size (required for `fit="fill"` to be a 1:1 transform).
- Replaced the three `hit_region` tags with a `MouseEvent(x, y)` handler that does grid/rect math in canvas coordinate space, using the canvas-click `MouseEvent` capability landed by an earlier host change — this is the payoff that lets clicking work again, alongside the existing keyboard nav (arrows + 1-9 + n), which is unchanged.
- Drew borders as a separate outline `CanvasRect` behind an inset fill rect instead of the removed `border_color`/`border_width` kwargs.
- Ported `apps/dev/canvas-sidebar` (identical crash) the same way — including fixing `HStack.to_node()`, which emitted a dead `"stack"` node type no renderer ever consumed.
- Fixed the stale `src/protocol/events.rs` doc comment claiming the SDK exposes `canvas_width`.

## Test Evidence

- `just test`: **1416 passed; 0 failed; 2 ignored**
- `mypy sdk/python/plexi_sdk/ --ignore-missing-imports --check-untyped-defs --exclude testing.py`: **no issues found in 26 source files**
- `plexi-alpha app test apps/sudoku`: **9 passed**
- `just check-capability-docs`, `just check-authoring-docs`, `just check-sdk-docs`: all pass
- **Live verification on alpha/PR channel** (rebuilt from source each time, not overlay hacks): sudoku renders correctly on both menu and game screens (`semantic.nodes` populated, `guest_fps` > 0, no decode errors), difficulty buttons render with correct border+fill, keyboard nav confirmed (arrow key changes difficulty selection, space starts a game). Canvas-sidebar renders 8 real nodes live via `just pr-install 2405`, zero decode errors.
[...]

PR #2408, opened 2026-07-16
fix(sdk): Tabs/Grid/Toggle/ProgressBar emit decoder-unsupported node types (stint 0402)

ianjamesburke:
No linked GitHub issue — tracked as stint task 0402.

## Summary

`Tabs`, `Grid`, `Toggle`, and `ProgressBar` in `sdk/python/plexi_sdk/ui.py` all
composed via `{"type": "stack"}` — the same dead type `HStack` emitted before
stint 0394 fixed it (PR #2405) — and `Tabs`/`Toggle` additionally wrapped
content in `{"type": "interactive"}`. Neither `"stack"` nor `"interactive"`
has a decode arm in `decode_node_data` (`src/host/wasm_python.rs`), and
`"interactive"` isn't in the WIT `ui-node-data` variant (`wit/plexi.wit`) at
all — confirmed by reading both files directly, not by grep. Both crash any
live CPython-WASM app that renders these widgets with `unsupported
CPython-WASM UINode type: <name>`, silent until run live.

- **Tabs/Grid**: compose from `row`/`column` instead of `stack`.
- **Tabs/Toggle**: tab and toggle controls are now `Button` nodes (`label` +
  `on_click` + `style`) instead of `Interactive`-wrapped text — `Interactive`
  isn't a real node type, and `Button` already covers the click-target need.
- **ProgressBar**: emits the decoder's existing native `progress-bar` arm
  instead of hand-rendering block-character text. Drops the `color`
  constructor param, which the decoder has never read from JSON (dead since
  introduction — confirmed by reading the decode arm).
- Adds a drift-guard test (`test_v3_5_uinode_widgets_emit_decoder_supported_root_type`)
  asserting every root type these four widgets can emit is in the decoder's
  supported set, so this class of bug can't silently reach a live host again
  for this widget set.
- Fixes `test_hstack_sized.py`'s stale assertion — it still expected `"stack"`
  after 0394 already changed `HStack` to emit `"row"`.

**Scope note:** no maintained/core app used these four widgets (confirmed via
constructor-call grep, not just import grep — `component-gallery` imports
`TabBar`/`Progress`, a separate older widget set, not `Tabs`/`ProgressBar`).
`Sized`, `Section`, `ActionBar`, and `Clickable` remain unfixed — no live
consumer today (`Sized`/`Section`'s only real usage was already removed in
0394; `ActionBar`/`Clickable` are untouched, dev-only via
`apps/dev/component-gallery`). `ActionBar` in particular emits `"action_bar"`,
also with no decoder or WIT support — a real, separate gap worth its own
stint task since it's not marked as the newer PGAP v3.5 widget set the other
four are.

## Test evidence

- `uv run pytest sdk/python` (from `sdk/python/`): **149 passed, 1 skipped**
  (full suite, unrelated to just the touched files).
- `uv run mypy plexi_sdk/ui.py tests/test_new_components.py`: 5 pre-existing
  errors, identical count/lines on `HEAD~1` — zero new errors introduced.
- `cargo test --bin plexi`: **1420 passed, 0 failed, 2 ignored** (one
  unrelated `assistant::` test flaked once under full-suite parallelism,
  confirmed to pass in isolation and on a clean re-run of the full suite).
- `just check-sdk-docs` / `check-authoring-docs` / `check-cli-docs` /
  `check-config-docs` / `check-capability-docs` / `check-docs-coverage`: all
  green (`sdk.md` regenerated via `just gen-sdk-docs` for the two changed
  docstrings).
- Live verification: built `target/debug/plexi` directly from this worktree
  (the installed `plexi-alpha` binary's baked-in `CARGO_MANIFEST_DIR` points
  at an already-deleted worktree from a merged stint, so `plexi-alpha`/`app
  render`/`app check` all hang — an unrelated, pre-existing environment
  issue, reproduced on an untouched app). Live GUI verification via
  `just pr-install <this PR>` still pending — will confirm in a follow-up
  comment.

## Do not merge

Per request — open for review only.

### DIFF
--- a/sdk/python/plexi_sdk/ui.py
+++ b/sdk/python/plexi_sdk/ui.py
@@ -2325,20 +2325,15 @@ class Tabs:
         for i, (label, _content) in enumerate(self.tabs):
             bold = i == self.active
             tab_buttons.append({
-                "type": "interactive",
-                "node_id": f"tab_{i}",
-                "child": {
-                    "type": "text",
-                    "text": label,
-                    "bold": bold,
-                },
-                "on_click": True,
-                "on_hover": False,
+                "type": "button",
+                "label": label,
+                "on_click": f"tab_{i}",
+                "style": "primary" if bold else "secondary",
+                "disabled": False,
             })
 
         tab_bar: dict = {
-            "type": "stack",
-            "direction": "horizontal",
+            "type": "row",
             "children": tab_buttons,
             "gap": SPACE_SM,
         }
@@ -2354,8 +2349,7 @@ class Tabs:
             children.append(active_content)
 
         return {
-            "type": "stack",
-            "direction": "vertical",
+            "type": "column",
             "children": children,
             "gap": 0.0,
         }
@@ -2405,15 +2399,13 @@ class Grid:
             for child in row_items:
                 row_nodes.append(child.to_node() if hasattr(child, "to_node") else child)
             rows.append({
-                "type": "stack",
-                "direction": "horizontal",
+                "type": "row",
                 "children": row_nodes,
                 "gap": self.gap,
             })
 
         return {
-            "type": "stack",
-            "direction": "vertical",
+            "type": "column",
             "children": rows,
             "gap": self.gap,
         }
@@ -2422,7 +2414,8 @@ class Grid:
 class Toggle:
     """On/off toggle switch (L1 sugar).
 
-    Renders as an Interactive node with a horizontal stack indicator.
+    Renders as a Button whose label carries the on/off state; clicking it
+    fires `node_id` as the on_click handler.
 
     Example::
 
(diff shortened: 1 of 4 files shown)


ianjamesburke:
Live verification complete via \`just pr-install 2408\` → \`plexi-pr-2408\`.

Built a throwaway global-scope probe app (\`Column\` doesn't accept these widgets as children — \`Tabs\`/\`Grid\`/\`Toggle\`/\`ProgressBar\` aren't \`Component\` subclasses — so each was returned directly as the \`view()\` root, cycling on spacebar) and drove it through a real booted host:

- **Tabs** → \`column > [row > [button "Tab 1", button "Tab 2"], text "Panel A"]\`
- **Grid** → \`column > [row["a","b"], row["c","d"]]\`
- **Toggle** → \`row > [text "Dark mode", button "on"]\`
- **ProgressBar** → native \`progress_bar\` node, \`value: 0.5\`

Zero \`unsupported CPython-WASM UINode type\` errors in \`~/.plexi-pr-2408/plexi.log\` across all four. Confirms the fix live, not just via the Python unit suite.

Side note unrelated to this PR: the previously-installed \`plexi-alpha\` binary's baked-in \`CARGO_MANIFEST_DIR\` pointed at a worktree from an already-merged/deleted stint (0400-0403), which made \`plexi-alpha\`, \`plexi app render\`, and \`plexi app check\` all hang indefinitely (reproduced on an untouched app, \`apps/dev/counter-tree\`) — not something this PR should fix, but worth a fresh \`plexi-alpha\` reinstall.

PR #2418, opened 2026-07-17
fix(assistant): SDK context/tool-cap pause, tool trail collapse, composer centering, drop header

ianjamesburke:
## Summary

Bundles four small assistant stints into one PR, per the 0413 gate follow-up work.

- **0422** — Inject the generated SDK API reference (`website/src/content/docs/sdk.md`, generated by `just gen-sdk-docs`) into the assistant's system prompt when the `build-plexi-app` skill is active, so it stops exploring `plexi_sdk` via PTY reads. Raise `MAX_TOOL_ITERATIONS` for those app-build turns, and make hitting the cap a graceful pause: one more tool-free model round summarizes progress instead of force-stopping mid-build on a canned stub.
- **0416** — Collapse consecutive successful tool-call rows into a compact trail instead of one row per call. Grouping is a pure function (`model::group_turns_for_render`) so it's unit-testable without egui. Failures and permission denials always stay visible on their own row. Added a regression test locking chronological order for interleaved tool calls + a queued follow-up message.
- **0423** — Vertically center the composer's hint text and caret (they were top-aligning against different row heights due to mismatched hint/body font sizes) and add left padding so the caret isn't flush against the hint text.
- **0424** — Delete `draw_header` and its `HEADER_BAR_H`/`HEADER_FONT_SIZE` constants; the transcript now starts at the top of the pane. No control in the bar was load-bearing (read-only labels only), so nothing needed relocating. Updated scenes/tests that asserted on the removed header label to check the still-present footer hint-bar label instead.

Stints: 0422, 0416, 0423, 0424 (no linked GitHub issues).

## Test plan

- [x] `cargo build` — clean
- [x] `cargo test --bin plexi` — full suite green: **1454 passed; 0 failed; 2 ignored**
- [x] `cargo test --bin plexi scenes::tests::scene_suite -- --exact` — all scenes pass, including updated `assistant-composer`, `assistant-multi-pane`, `assistant-settings`
- [x] New/updated coverage: `broker::tests::tool_loop_pauses_gracefully_after_max_iterations`, `skills::tests::app_build_sdk_reference_is_generated_and_nonempty`, `model::tests::group_turns_for_render_collapses_only_consecutive_successful_tool_calls`, `model::tests::tool_calls_and_queued_message_stay_in_chronological_order`
- [x] Visual evidence: `assistant-composer-empty.png` scene shot confirms centered hint/caret with left padding
- [ ] Binary install validation (`just pr-install <PR>`) — recommended for the composer/header visual changes; not run here

Not merging — a separate tester validates first.


ianjamesburke:
Fixed in 3ef6bcb6.

Root cause confirmed: `SkillRegistry::matching_enabled`'s 50%-word-overlap bar scored the tester's prompt ("Build me a tiny counter app with plus and minus buttons") at 2/6 = 0.33 against the skill's own description, so the app-build skill silently never loaded — meaning neither the SDK reference injection nor the raised tool-call cap ever activated (both gate on the same `selected_skill`).

Added `SkillRegistry::app_build_fallback`: an independent build-verb + build-noun check (word-boundary tokenized) that activates the app-build skill whenever the general matcher misses a plausible build request. Wired in as a fallback after `matching_enabled` in `start_turn`.

Verification:
- New regression tests confirm the exact tester prompt: `skills::tests::app_build_fallback_catches_prompt_the_general_matcher_misses` (documents the 2/6=0.33 miss), `assistant::tests::app_build_prompt_missed_by_general_matcher_still_gets_sdk_context_and_raised_cap` (end-to-end via a capturing broker — asserts the actual outgoing request has the SDK reference in `system` and `max_tool_iterations == Some(60)`).
- Installed `plexi-pr-2418` and drove the real Assistant pane with the tester's exact prompt: log confirms `assistant[...]: loading builtin skill 'build-plexi-app' from <builtin>` (previously silent). Full end-to-end completion wasn't observable in my automated environment — a fresh app-signature keychain authorization prompt blocked the OpenRouter key lookup — but skill activation plus the deterministic request-inspection test together confirm the fix.
- `cargo test --bin plexi`: **1457 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 81.57s**

PR #2484, opened 2026-07-24
fix(input): deliver IPC text to raw-focus panes; rename inputs pre-select contents (stints 0537, 0545)

ianjamesburke:
Batch PR for stint tasks **0537** and **0545** (no linked GitHub issues). Both root-caused to the stint-0429 focus reconciler's post-frame focus grants.

## Stint 0537 — fix pre-existing alpha test failure: `send_to_app_pane_injects_text_through_focused_render_input`

**The fix was to the code, not the test.** The defer-until-focused gate added in #2467 (`raw_input_hook`) only released queued IPC text when the egui-focused id was *registered* in the focus surface registry. A widget holding raw egui focus (the exact "misbehaving widget" pattern the reconciler tolerates by design, and what the probe simulates) never registers, so `plexi pane send` text deferred forever — the test predates the gate and asserts a real invariant.

Fix: new `input_owner::pane_receives_ipc_text` — text is deliverable when the focused id is the pane's registered surface, **or** an id unknown to the registry *and* the pane already owned input on the last reconciled frame. The owner-history guard (found by pre-push Codex review) prevents a just-switched focus target's text from being swallowed by the previous pane's raw-focus widget. With nothing focused (pane created during hidden passes), text still waits — the case the gate was built for.

Test-side adjustment (secondary): `TextInputProbe` now grabs raw focus once instead of every frame — a per-frame re-grab across two probe panes flips egui focus mid-pass so no TextEdit ever draws while focused, which is a pathology of the probe, not the host. It also takes an explicit TextEdit id so multi-probe tests don't collide on auto ids.

## Stint 0545 — rename inputs pre-select their contents

User-confirmed live: Cmd+R / Cmd+Shift+R opened with caret at end, nothing selected, despite `select_all_on_focus(true)` at every call site. Root cause: the reconciler grants focus via `request_focus` **after** every widget has drawn, so by the widget's next pass egui's `id_previous_frame` already names it and `Response::gained_focus()` never returns true — `select_all_on_focus` was dead code for all reconciler-granted focus.

Fix in the shared widget (`src/ui/text_field.rs`), not per call site: `TextField`/`TextArea` track the focus state observed on their own previous pass and fire select-all on that edge. Also fixes a latent byte-vs-char index bug (`buf.len()` → `chars().count()`) in the selection range. File Browser's rename modal now opts in (`.select_all_on_focus(true)`), which it never did.

## Tests

- `send_to_app_pane_injects_text_through_focused_render_input` — pre-existing, now green (was deterministically red on clean alpha).
- `send_to_pane_never_routes_text_into_previous_panes_raw_focus_widget` — new; misroute guard.
- `pane_receives_ipc_text_accepts_registered_and_raw_focus` — new unit test for the gate.
- `rename_pane_overlay_preselects_prefill`, `rename_context_overlay_preselects_prefill`, `file_browser_rename_modal_preselects_entry_name` — new; all three written red first (reproduced the live report exactly: caret at end, empty selection), green after the fix. File Browser test drives the real `KeyPane cmd+r` path.

**Test evidence:**
- cargo test (env-stripped): **1823 passed, 0 failed, 2 ignored** — full bin suite.
- Diff classification: host logic + shared text-widget behavior; no layout/geometry/style change, selection state asserted directly by harness tests (return-value/invariant → Rust `#[test]` per TESTING.md).
- Codex review: run 1 flagged one P1 (raw-focus misroute on focus switch) — fixed with the owner-history guard + regression test; run 2 clean.
- Conclusion: **binary install required** — 0545 is a live user-reported interaction (Cmd+R / Cmd+Shift+R feel); validator should `just pr-install <PR>` and confirm rename fields open fully selected in File Browser, Cmd+R, and Cmd+Shift+R.

No linked GitHub issues; stint tasks 0537 and 0545 are the trackers.

### DIFF
--- a/src/app/input_owner.rs
+++ b/src/app/input_owner.rs
@@ -218,6 +218,7 @@ impl PlexiApp {
         }
 
         store_previous_surfaces(ctx, registry);
+        store_previous_owner(ctx, owner);
     }
 }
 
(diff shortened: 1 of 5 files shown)


ianjamesburke:
## Fix round (tester reject: 0545 not fixed in real build)

**Root cause of the divergence:** `Response::has_focus()` is `i.focused && memory.has_focus(id)` — gated on **OS window focus**. Every CLI-driven session (`plexi pane send/key`, drive-host, the tester) runs with the host window blurred, so the previous focus-edge detector read `false` for the entire session and select-all never fired in the real build. The harness passed because kittest's `RawInput::default()` has `focused: true`. The reconciler deliberately ignores OS focus for exactly this injection path; the edge detector now reads pure `memory.has_focus` to match.

**Live verification against a locally built host, driven blurred via CLI** (`~/.plexi-fixtest/plexi.log`):

```
18:34:11 [INFO] [plexi::ui::text_field] file_browser_rename: select-all on focus gain
18:34:13 [INFO] [plexi::file_browser] file_browser: renamed '.../fb-test/beta_two.txt' to '.../fb-test/QQ'
```

Typed `QQ` over the open rename field → file renamed to `QQ` (replaced, not appended), **in the source's own directory**.

**Separate bug (rename relocated the file to host cwd): fixed in the same file, one construct** — target is now `source.parent().join(name)` instead of `self.cwd.join(name)`, with an explicit error when the source has no parent.

**Harness hardening:** new `run_blurred_frames` helper; the FB preselect test now runs OS-blurred frames (modeling the CLI-driven session that the first fix silently failed in) and asserts the full flow end-to-end: selection spans the prefill → typed text replaces it → Enter renames in place (`QQ` exists, `beta_two.txt` gone, `beta_two.txtQQ` never created).

Suite: `cargo test --bin plexi` (env-stripped) → **1823 passed, 0 failed, 2 ignored**. `cargo build --bin plexi` → clean (pre-existing `__eh_frame` linker note only).

PR #2489, opened 2026-07-25
perf(wasm): coalesce decoder wakes, shrink the canvas wire payload, pick the smaller tree encoding (stints 0549, 0550, 0551)

ianjamesburke:
Lands stints **0549** (coalesce the WASM decoder frame-arrival wake), **0550** (shrink the canvas wire payload), and **0551** (skip the tree_delta when it is not smaller) in one PR — they all sit on the shared CPython-WASM canvas wire/render path and were all found by the 0541 balls audit. No linked GitHub issues; stint is authoritative.

## What changed

**0549 — `src/host/wasm_python.rs`.** Two independent repaint sources fired per guest frame: `PythonFrameScheduler` scheduled the next admission deadline from `ui()`, and `PythonOutputDecoder::decode_loop` separately called `request_repaint_after_for(1ns)` on every produced frame. `ui()` now publishes the repaint it just scheduled (`ScheduledWake`) to the decoder thread, and `decoder_wake_needed` lets a frame ride that paint instead of forcing a second full-host repaint. The `1ns`-not-zero convention is preserved wherever the wake is kept. Suppressed wakes are drained into the periodic `CPython-WASM perf` line as `suppressed_wakes=N`.

Suppression is deliberately narrow — it applies only when **all** of these hold:
- the app is in continuous mode *and* the frame scheduler itself has a deadline pending (a full pipeline publishes no cadence tag, because the next admission depends on this very frame arriving);
- the whole produced batch is frame cadence output — the tree plus its `frame_done`. A host command, or a decode error, still wakes immediately (`is_frame_cadence_output`);
- the published schedule is not stale. `eframe` skips `ui()` entirely while the window is hidden, so a wake overdue by more than one interval is no longer evidence a paint is coming, and the decoder wakes.

**0550 — `sdk/python/plexi_sdk/ui.py`, `_v3_runtime.py`.** Canvas geometry is rounded at the source through `_wire()` under the `CANVAS_WIRE_DECIMALS` policy constant (not an inline magic number), and the emit path serialises with compact separators via `_dump_line`. Both are schema-preserving — floats stay floats (`float()` before `round()`, so an int argument does not become an int on the wire), no key changes — so **the host decoder is untouched**.

**0551 — `sdk/python/plexi_sdk/_v3_runtime.py`.** `_emit_tree` now serialises both candidate encodings and emits the smaller one per frame, instead of assuming the delta always wins. The fallback is the existing `component_tree` shape — no new wire shape — and the fail-loud `apply_delta` / host-decoder contract is untouched.

## Measured, before and after

Reproduced first, per the Issue Visibility rule. Guest-side numbers are from the real `apps/balls` view at 50 balls, encoded through `_encode_uitree` + `node_patch` in this worktree (baseline via `PLEXI_SDK_PATH` pointed at an unmodified `HEAD` SDK export).

| 50-ball balls frame | before | after |
|---|---|---|
| full frame on the wire | 17.46 KiB | **10.63 KiB** (61%) |
| `json.dumps` | 0.249 ms | **0.158 ms** (63%) |
| encoding chosen | `tree_delta`, 17.93 KiB (2.7% *larger* than the full frame) | `component_tree`, 10.63 KiB |

Isolating the two 0550 reductions: rounding alone 12.15 KiB (70% of baseline), compact separators alone 15.93 KiB (91%), both together 10.63 KiB (61%). At 60 fps that is ~1048 KiB/s of stdout down to ~638 KiB/s.

**0551 verified in both directions** on a 50-command canvas with a varying number of commands moving per frame:

| commands moving | delta | full frame | emitted |
|---|---|---|---|
| 1/50 | 0.15 KiB | 3.48 KiB | delta |
| 5/50 | 0.43 KiB | 3.49 KiB | delta |
| 25/50 | 1.89 KiB | 3.53 KiB | delta |
| 50/50 | 3.72 KiB | 3.58 KiB | **full frame** |

Cost of making the choice exact: when the delta wins, the full frame is serialised and discarded — 0.039–0.044 ms/frame, i.e. 0.26% of a 16.7 ms frame budget. I took the exact comparison over a size heuristic; the measured overhead is small enough that a fudge factor would not have earned its ambiguity.

`paint_fps` / `guest_fps` / `avg_roundtrip_ms` on a live host are **not** measured here — a worker pane must not boot or drive the GUI host. That is the tester's job; drive steps below.

## The two open questions, resolved

**0549: is the latency cost imperceptible for input-driven apps?** Yes, because input-driven apps never suppress at all. Scheduled-mode apps publish no continuous interval, so `decoder_wake_needed` always returns true for them — asserted directly (`decoder_always_wakes_for_a_scheduled_mode_app`) and end to end against the real `apps/dev/key-event-probe` subprocess, which must show exactly zero suppressed wakes after a real `KeyPane` round trip (`scheduled_mode_app_never_suppresses_its_frame_wake`). For an app that *is* driving a continuous cadence, the ceiling is one frame interval (≤16.7 ms at 60 fps) and typically far less, since the frame arrives near its own deadline; those apps are animation-driven by definition, and the trade buys back a full host repaint — all panes and chrome — per guest frame.

**0550: does any canvas usage need finer than 2dp?** No. What I checked:
- Every `Canvas(...)` construction in the repo. Every one declares a pixel-scale logical space: either the live pane size (`fit="fill"` — `balls`, `sudoku`, `breakout`, `canvas-sidebar`) or a fixed 360–640 unit box (`fit="contain"` — `tetris`, `snake`, `canvas-grid`, `canvas-click-probe`). No normalised (0..1) space anywhere.
- The host magnification path, `canvas_transform` in `src/host/wasm_render.rs`: `fill` scales by `rect/logical` (≈1, since the app passes the pane size); `contain` uses a uniform `min(sx, sy)`. Worst realistic case is a fixed 360×440 canvas full-screen on a 2560×1440 display — 3.27×. Half a rounding step is then 0.016 logical px, or 0.033 device px on a 2× display. A visible ≥1-device-pixel error would need roughly 200× magnification, i.e. a canvas under ~10 logical units.
- Pixels, not just numbers: headless `plexi app render --png` of `sudoku` and `snake` before and after is **byte-identical**. (`tetris` differs, but it is nondeterministic — two renders on the *same* SDK also differ, so it proves nothing either way.)
- The bound is pinned by a test (`test_rounding_is_below_one_device_pixel_at_realistic_scales`) so the policy cannot silently drift.

The precision is fixed as `CANVAS_WIRE_DECIMALS`, documented as a wire policy including the constraint that it assumes a pixel-scale coordinate space.

## Review finding considered and declined

Codex flagged (P2) that a future normalised canvas (`width=1`, coordinates ~0.004) would quantise to `0.0`, and suggested the constructor reject or opt out of such a space. I did not add that guard: `fit="fill"` apps pass the **live pane size** straight into `Canvas(width=…, height=…)`, so any minimum-dimension check would raise from inside `view()` the moment a pane rect is transiently collapsed or sub-unit — a real regression for existing apps in exchange for a guard on an app that does not exist. The constraint is documented on the policy constant instead. Worth revisiting as its own task if a normalised coordinate space is ever actually wanted.

## Test evidence

- `cargo build --bin plexi` — green.
- `cargo test --bin plexi` (env-stripped) — **1857 passed; 0 failed; 3 ignored**. Baseline before the two new harness tests was 1855.
- `pytest sdk/python/tests apps/balls/tests` — **256 passed, 1 skipped**. One deselect: `test_version_unification.py::test_dunder_version_matches_pyproject`, which **fails identically on clean alpha** (`__version__` 3.0.0 vs pyproject 0.2.0) and is unrelated to this diff.
- `mypy sdk/python/plexi_sdk/` — Success, no issues in 27 source files.
- `just check-sdk-docs` — up to date, no regeneration needed.
- Memory watchdog ran alongside both suites (kills any `plexi-*` test process over ~8 GB RSS); it never fired.
- Visual: `snake` headless render read and confirmed to show real seeded content, identical before and after.

New tests: `decoder_wakes_when_nothing_is_scheduled`, `decoder_always_wakes_for_a_scheduled_mode_app`, `decoder_rides_a_continuous_paint_already_within_one_frame`, `decoder_wakes_when_the_continuous_paint_is_more_than_a_frame_away`, `decoder_wakes_when_the_published_schedule_has_gone_stale`, `frame_done_rides_the_cadence_but_other_messages_do_not`, `continuous_interval_is_reported_only_in_continuous_mode`, `a_full_continuous_pipeline_has_no_render_deadline_to_ride`, `continuous_app_frames_ride_the_scheduled_repaint`, `scheduled_mode_app_never_suppresses_its_frame_wake`, plus `sdk/python/tests/test_canvas_wire.py` and two delta-direction tests in `test_v3_delta.py`.

**Conclusion: binary install required.** Tier 2 — this touches the WASM render container, and the headline `paint_fps` claim can only be confirmed against a live host.

## Drive steps for the tester

```bash
just pr-install <PR>
```

1. Launch balls on the PR build: `plexi-pr-<PR> app launch apps/balls` (or from the app menu). Leave it running ~30 s.
2. Read the telemetry: `grep "CPython-WASM perf" ~/.plexi-pr-<PR>/plexi.log | tail -20`.
   - **Expected:** `paint_fps` converges toward `guest_fps` (~60 rather than ~75), `guest_fps` unchanged at ~60, `stdout_kib` roughly 60% of its old value, and `suppressed_wakes` a nonzero count each interval.
   - Compare against the same run on `plexi-alpha` for the before numbers.
3. Confirm balls still looks right — smooth motion, no stutter, no visual quantisation of ball positions.
4. Input latency check (this is the 0549 risk surface): open an input-driven app — `apps/dev/key-event-probe`, plus any Core app with text entry — and confirm keystrokes and clicks feel exactly as immediate as on alpha. That path is asserted to never suppress; this is the human confirmation.
5. Run `tetris` or `snake` (`fit="contain"`, i.e. the magnified coordinate path) full-screen and confirm no visible geometry shift or jitter versus alpha.

### DIFF
--- a/sdk/python/plexi_sdk/_v3_runtime.py
+++ b/sdk/python/plexi_sdk/_v3_runtime.py
@@ -27,15 +27,29 @@ from plexi_sdk._v3_state import StateSnapshot
 
 
 _LOCK = threading.Lock()
-_EMIT_BATCH: list[dict] | None = None
+_EMIT_BATCH: list[str] | None = None
+
+#: Protocol output carries no human reader, so the default `", "` / `": "`
+#: separators are pure wire weight on every key of every frame.
+_JSON_SEPARATORS = (",", ":")
+
+
+def _dump_line(obj: dict) -> str:
+    """Serialise one protocol message to its exact wire line."""
+    return json.dumps(obj, separators=_JSON_SEPARATORS) + "\n"
 
 
 def _emit(obj: dict) -> None:
+    _emit_line(_dump_line(obj))
+
+
+def _emit_line(line: str) -> None:
+    """Write an already-serialised protocol line, or queue it in the batch."""
     with _LOCK:
         if _EMIT_BATCH is not None:
-            _EMIT_BATCH.append(obj)
+            _EMIT_BATCH.append(line)
             return
-        sys.stdout.write(json.dumps(obj) + "\n")
+        sys.stdout.write(line)
         sys.stdout.flush()
 
 
@@ -54,7 +68,7 @@ def _finish_emit_batch() -> None:
         _EMIT_BATCH = None
         if not batch:
             return
-        sys.stdout.write("".join(json.dumps(obj) + "\n" for obj in batch))
+        sys.stdout.write("".join(batch))
         sys.stdout.flush()
 
 
@@ -349,29 +363,41 @@ class V3AppRuntime:
         the host never has to guess what a patch applies to. Relies on the
         documented `view()` purity contract: nodes are rebuilt every frame,
         never cached and mutated in place across frames.
+
+        A delta is only a win when it is actually smaller than the frame it
+        replaces. On a full-motion canvas, where every command changes every
+        frame, each changed command is re-sent wrapped in its own index — so
+        the delta comes out slightly *larger* than the full frame. Both
+        encodings are already supported by the host decoder, so the smaller of
+        the two wins on a per-frame basis rather than by assumption.
         """
         prev = self._last_tree
         self._last_tree = encoded
         nodes = encoded["nodes"]
         force_full = self._force_full_tree
         self._force_full_tree = False
+        full = {"type": "component_tree", "frame_id": frame_id, "tree": encoded}
         if (
             force_full
             or prev is None
             or encoded["root"] != prev["root"]
             or len(nodes) != len(prev["nodes"])
         ):
-            _emit({"type": "component_tree", "frame_id": frame_id, "tree": encoded})
+            _emit(full)
             return
         changed: list[dict] = []
         for node, old in zip(nodes, prev["nodes"]):
             if node == old:
                 continue
             if node["key"] != old["key"]:
-                _emit({"type": "component_tree", "frame_id": frame_id, "tree": encoded})
+                _emit(full)
                 return
             changed.append(_node_patch(node, old))
-        _emit({"type": "tree_delta", "frame_id": frame_id, "changed": changed})
+        delta_line = _dump_line(
+            {"type": "tree_delta", "frame_id": frame_id, "changed": changed}
+        )
+        full_line = _dump_line(full)
+        _emit_line(delta_line if len(delta_line) < len(full_line) else full_line)
 
     def _handle_key(self, ev: dict, *, schedule_render: bool = True) -> None:
         key = _normalize_key(ev.get("key", ""))
--- a/sdk/python/plexi_sdk/_v3_runtime.py
+++ b/sdk/python/plexi_sdk/_v3_runtime.py
@@ -33,6 +33,16 @@ _EMIT_BATCH: list[str] | None = None
 #: separators are pure wire weight on every key of every frame.
 _JSON_SEPARATORS = (",", ":")
 
+#: How many delta-eligible frames ride a previously measured full-vs-delta
+#: verdict before it is re-measured. Measuring means serialising the frame
+#: both ways, which costs more guest CPU than the byte difference it can
+#: recover — on a full-motion canvas the double serialisation alone measurably
+#: lowered achievable guest_fps. The winner is a property of the app's frame
+#: shape, not of any single frame, so it is stable between structural changes;
+#: every full frame resets the countdown so the first delta-eligible frame
+#: after one always re-measures.
+_TREE_SIZE_CHECK_INTERVAL = 30
+
 
 def _dump_line(obj: dict) -> str:
     """Serialise one protocol message to its exact wire line."""
@@ -99,6 +109,8 @@ class V3AppRuntime:
         self._last_render_time: float | None = None
         self._last_tree: dict | None = None
         self._force_full_tree = True
+        self._prefer_full_tree = False
+        self._tree_size_check_in = 0
         self._frame_id = 0
         self._app_id = ""
         self._workspace_root = ""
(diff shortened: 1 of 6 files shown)


ianjamesburke:
## Fix round: balls guest_fps regression (tester blocker)

Bisected the ~5% guest_fps loss to two guest-side CPU costs added by 0550/0551, together +72% on the per-frame emit path (native microbench of balls-shaped frames):

1. **0551's full-vs-delta byte compare serialised the whole tree twice every delta frame** (dominant, ~0.041ms of 0.139ms). Fixed in `7b37533f`: the verdict is measured (both encodings serialised) only on the first delta-eligible frame after any full frame and every 30 frames after, and remembered in between — the winner is a property of the app's frame shape, not any single frame. Steady-state frames serialise exactly one encoding.
2. **`round(float(v), 2)` per coordinate** (~0.017ms). Same 2dp quantisation now via multiply/floor-divide (~3.5x cheaper); tie direction is explicitly outside the wire policy.

### Re-measured on the installed build (head 7b37533f), ephemeral hosts, 25× 2s windows per run, A-B-A-B

| build | guest_fps (r1 / r2) | wire KiB/frame | paint_fps | suppressed_wakes |
|---|---|---|---|---|
| alpha | 55.6 / 55.7 | 3.70 / 3.69 | 110.6 | — |
| PR fixed | 55.1 / 55.5 | 2.25 / 2.29 | 55.1 (= guest) | ~111/interval |

guest_fps is back to alpha parity (≤1%, inside alpha's own run-to-run spread); the 0550 wire win (−39%) and 0549 paint coalescing both survive.

### Gates
- `cargo build --bin plexi`: green
- env-stripped `cargo test --bin plexi`: **1857 passed, 0 failed, 3 ignored**
- `uv run mypy plexi_sdk` (sdk/python): **Success: no issues found in 27 source files** (test-file mypy errors are pre-existing and byte-identical before/after)
- sdk pytest: **253 passed, 1 skipped**; `test_version_unification` fails identically on untouched alpha (pyproject 0.2.0 vs `__version__` 3.0.0 — pre-existing, unrelated)

New regression guards: `test_steady_state_delta_frames_serialise_at_most_once`, `test_full_frame_preference_persists_between_measures`.

PR #2499, opened 2026-07-27
feat(mcp): bridge configured MCP servers onto the connector tool plane (stint 0382)

ianjamesburke:
Closes stint **0382 — Assistant MCP connector support**. No linked GitHub issue.

## What this is

Plexi does not embed an MCP client. MCP servers are consumed through the existing app-connector plane: a bridge app hosts the client and re-exposes the server's tools via `ExposeTools` (stint 0369), so every MCP tool the assistant can reach is an ordinary connector tool — capability-scoped, permission-gated, visible in the event log. There is no second tool path.

## The split, and why

SDK apps run as CPython-in-Wasmtime with no ambient process access, so an app cannot spawn a server. More importantly it must not be able to: an app-supplied argv would be arbitrary native code execution wearing a capability's clothes.

- **The host owns the process** (`src/host/mcp_client.rs`): a user-owned `mcp_servers.toml` registry plus a stdio supervisor that spawns a configured server, frames newline-delimited JSON, captures stderr for diagnostics, and kills the child with its pane. It never parses a JSON-RPC body and does not know what `tools/list` means.
- **The app owns the protocol** (`apps/dev/mcp-bridge`): `initialize` → `notifications/initialized` → `tools/list` → schema translation → `ExposeTools`, then `ToolCall` → `tools/call` → `ToolResult`.

**Containment property:** an app names a server by id; the host resolves the argv from user config. `mcp.client` therefore authorizes *"talk to a server the user configured"*, never *"run this command"* — a compromised bridge app cannot reach a binary the user has not already listed. The wire shape is asserted by a test: `mcp_connect` carries only `request_id` and `server_id`, and if a command field ever appears there the capability has silently become process execution.

## Schema translation fails loudly

A tool whose `inputSchema` is missing, is not an object, has non-object `properties`, has no name, or duplicates an earlier name is **skipped with a logged reason** and listed in the pane. Never silently absent, never exposed with a guessed schema — an assistant calling a tool whose schema was invented fails far worse than one that can see the tool is unavailable. `$schema` draft keys are stripped; declared `outputSchema` is preserved verbatim, and tools without one declare MCP's real `content`-array reply envelope rather than a richer shape the server never promised.

## Fixtures are real, not imagined

Everything under `src/host/testdata/mcp/` is verbatim wire traffic captured from `@modelcontextprotocol/server-filesystem` (14 tools, draft-07 schemas) and `@modelcontextprotocol/server-everything` (13 tools, real `readOnlyHint` annotations, a declared `outputSchema`). Nothing was reconstructed from docs.

---

## ⚠️ Two things for Ian — surfaced, not decided

**1. `read_only` / host-vs-app tool mutability.** This walks straight into the open ruling from PR 2495. Real MCP servers advertise `annotations.readOnlyHint`, and the everything server marks most of its tools `readOnlyHint: true`. Honouring that would let an *external, less-trusted* tool source convert the host's default `Ask` into an automatic allow.

**This PR does not decide it.** The bridge simply never sets `read_only` on any tool: every MCP-derived tool is ask-gated, consistent with the 0554 PRM's "silence is never read as read-only". That is the conservative refusal to use the app-declared upgrade path, not a new trust model. Whether the host can ever establish mutability for a bridged tool remains yours to rule on. Guarded by `test_no_bridged_tool_is_ever_declared_read_only`.

**2. Grant granularity — same tension as stint 0381.** Exposed tools are named `mcp.<server>.<tool>`, so the server is named in the consent prompt and the event log, and two servers exposing `read_file` are two separately-approvable tools. But grants still key on actor + tool name + workspace, so this is *naming*, not resource binding — there is no per-server grant. Same parked item as 0381's network grants. Surfaced, not decided.

## Also worth knowing

- **CPython bridge apps only.** These messages live in `LivePythonPane` alongside `http_request`, and never reach the WIT `effect` variant, so Rust WASM component apps cannot open an MCP connection. I initially added the WIT variants and it broke 71 tests: extending `wit/plexi.wit` changes the component type, and the host rejects every prebuilt component compiled against the old world (`expected variant of 25 cases, found 22`). Adding the WASM path means rebuilding the CPython shim bundle and every `apps/wasm-poc/` fixture in the same change — a separate unit of work. Documented in the module header.
- **`apps/dev/mcp-renderer` is dead and I left it alone.** It is a `schema_version = 1` app doing raw `subprocess.Popen`, which the Wasmtime runtime cannot do, and it uses the pre-v3 `App`/`RenderContext` API. `registry/mcp/*.json` still routes `plexi app open --mcp` at it. Out of scope here; worth its own cleanup task.

## Round 2 — Codex review response (commits `3298d287`, `a6d297b1`, `471c1ff7`)

**3. Sandbox gap — PARKED, needs Ian's ruling.** The review's item 1 said the MCP server launch bypasses the 0554 seatbelt helper. **That helper does not exist in-tree.** The only `sandbox_init`/seatbelt references anywhere are in `docs/assistant-authority-model.md` (the 0554 PRM itself), which explicitly defers the profile: *"The sandbox profile for the worker is deliberately a separate follow-on."* There is no sandboxed-subprocess launch path to route through, and building one is architectural work outside this PR. Until it exists, an MCP server runs with ambient authority — bounded by the user having explicitly configured its command line in `mcp_servers.toml`, and by the process-group containment below, but not OS-sandboxed.

The other review findings were one family — an externally-controlled subprocess whose lifetime, output volume, and responsiveness were assumed benign — and were fixed structurally: `McpConnection` is now constructed under explicit `McpLimits`, and every externally-controlled behavior is bounded, with violation tearing the connection down:

- **Lifetime**: the child is spawned as its own process-group leader (`process_group(0)`, same idiom as `build_exec.rs`); teardown SIGKILLs the whole group, so `npx`/`uvx` descendants die with the launcher. Guard: `drop_kills_the_whole_process_group`.
- **Output**: stdout lines capped at 8 MiB (overflow closes the connection), stderr lines truncated at 8 KiB, and the pane-facing queue is a bounded `sync_channel` built by the one sanctioned constructor (`inbound_channel`) — a flooding server blocks on its own pipe, never in host memory. Guards: `stderr_flood_is_truncated_and_the_connection_survives`, `discard_line_skips_to_the_next_line_after_an_overflow`.
- **Input**: writes moved to a dedicated writer thread behind a bounded queue; `send` never blocks the host thread, and a server that stops reading stdin surfaces as a fast error the pane converts into a close. Guard: `send_fails_fast_when_the_server_stops_reading_stdin`.
- **Responsiveness**: host-side first-output deadline (120 s backstop; a server that never speaks is killed with a named reason — guards: `silent_server_is_closed_at_the_first_output_deadline`, `responsive_server_outlives_the_first_output_deadline`) plus an app-side 90 s handshake deadline in the bridge via the existing `SetTimer`/`TimerFired` SDK surface — a server stuck at `initializing` is failed visibly and disconnected, and the close event does not overwrite the diagnosis.
- **Docs rule**: `src/config/AGENTS.md`, `sdk/python/AGENTS.md`, `skills/plexi-cli/SKILL.md`, and `docs/CONFIG.md` updated in this PR per the new standing rule.

Residual, documented in the module header: a descendant that calls `setsid` escapes the process group — closing that fully requires the 0554 sandbox worker above.

**Pre-existing failure surfaced (not this PR's):** `sdk/python/tests/test_version_unification.py` fails on unmodified alpha too — `plexi_sdk.__version__` is `3.0.0` while `sdk/python/pyproject.toml` says `0.2.0` (drift from the `v0.2.0` release commit). Needs its own fix.

## Test evidence

- `cargo test --bin plexi` (round 2): **1910 passed, 0 failed, 3 ignored** — full bin suite, env-stripped, memory watchdog clean. One earlier run flaked on the known load-sensitive environmental failure from the alpha baseline; green on rerun. `mcp_client` module: **17/17**.
- SDK suite (`test_v3_adapter.py` + `test_v3_runtime_regression.py`): **75 passed, 1 skipped**. Adds 6 MCP protocol tests. (Full `sdk/python/tests/` sweep additionally hits the pre-existing version-drift failure noted above.)
- `apps/dev/mcp-bridge/tests` (round 2): **24 passed** — handshake, deadline, translation against real captured schemas, proxying, JSON-RPC errors, server loss.
- `mypy sdk/python/plexi_sdk/`: **no issues in 27 source files**.
- `just check-docs`: all green (SDK docs regenerated).
- `plexi app check apps/dev/mcp-bridge`: **passed** at PR-open (manifest, SDK, mypy, 4 render sizes). Round 2 rerun against the *alpha profile* fails on import because the installed alpha SDK predates this branch's `Mcp*` effects — environmental; the `pr-install` profile carries the branch SDK.
- **End-to-end against a live server:** spawned real `@modelcontextprotocol/server-filesystem`, ran the actual bridge module over it — 14 tools exposed, 0 skipped, all `read_only=False`; called `mcp.filesystem.read_text_file` as the assistant would and got the real file contents back on the originating `call_id`.

Conclusion: **binary install required** for tester validation — the transport spawns real processes and the connector registration is live-host behavior. Validate with `just pr-install <PR>`, configure a server in `~/.plexi-pr-<N>/mcp_servers.toml`, then `plexi-pr-<N> app open apps/dev/mcp-bridge -- <server-id>`.

## Gate

Headless only, per the worker contract. No GUI host was installed or driven.


ianjamesburke:
Fix round pushed: 5841bfe6. One-line root cause per bug:

- **Bug 1 (host-kill flood):** the 16-slot channel bounded only the reader hop — `drain_runtime` shoveled every line into the guest's *unbounded* stdin `VecDeque`. Fix: guest stdin is now the enforced boundary (64 MiB hard cap, all-or-nothing push) with a watermark-gated pump (8 MiB) so backpressure propagates all the way to the server's own pipe.
- **Bug 2 (healthy servers killed at 90 s):** all MCP servicing lived only in the paint path (`LivePythonPane::ui`) — the documented hidden-window trap; an idle/occluded pane never drained either direction. Fix: Python pane runtime I/O serviced from `App::logic` (`service_python_pane_runtimes`) plus a per-line wake hook (`McpWake`). Deadlines untouched.
- **Bug 3 (11 MB/min log growth):** stderr line *length* was capped but rate was not; every line became a plexi.log write. Fix: windowed log budget (20 lines/10 s) with one rate-limited summary naming the suppressed count; `last_stderr` retention unchanged.

Headless guards: stdin hard-cap unit test, a real-flooding-server pump test against a guest that never reads, stderr budget unit tests, and `mcp_reply_reaches_guest_while_window_hidden` (new `apps/dev/mcp-hidden-probe` fixture driven by `hidden_frame` passes only — verified to FAIL with the logic-pass servicing removed). `cargo test --bin plexi`: 1916 passed, 0 failed.

PR #2510, opened 2026-07-28
feat(cli): agent orchestration verbs — pane send --submit, pane new --agent, pane slot wait (stints 0583, 0584, 0585)

ianjamesburke:
## Stints

- **0583** — `plexi pane send --submit`: one verb that types, settles, submits, confirms, and self-heals the collapsed paste with exactly one internal retry. Exit 0 = confirmed submitted; exit 1 = typed but unconfirmed with the observed input line on stderr. Settle/confirm is host-side off the PTY event stream (`last_pty_output_at`) and the rendered grid tail — explicitly not `pane capture` (task 0385's unreliability is what this verb routes around). Plain `pane send` and `pane command` stay type-only.
- **0584** — `plexi pane new --agent <tier-alias> [--boot-timeout <SECS>]`: spawns, launches the alias verbatim, and blocks until the agent's hook-reported state reaches idle (`SetAgentState` → `Pane::agent()` — the existing detector, no parallel PTY prompt scraper). Pane id on stdout only when ready; boot timeout = exit 2 with the created pane id on stderr so the caller can close it; never queues (requires PLEXI_SOCKET).
- **0585** — `plexi pane slot wait <name> [pane_id] --until <REGEX> --timeout <SECS>`: host-side, event-driven blocking read off the slot write path. Level-triggered (already-matching value returns immediately). Exit 0 = match on stdout, 2 = timeout, 1 = usage/plumbing.

All three waits park host-side in `src/app/pane_wait.rs` and are serviced from `App::logic`, so they complete and expire while the window is hidden/occluded (per-verb `*_while_window_hidden` harness guards).

## Provenance — salvaged implementation

The implementation predates this PR's author session: worker-13 implemented all three stints and its host crashed before committing. The staged index was salvaged intact (23 files), audited against the three stint bodies and the batch's architectural constraints (babysitter SKILL.md deletion ban — deprecation notes only; detector reuse; no capture dependency), and the only crash-lost piece was one line of generated docs (`pgap.md`'s `submit` field row), restored by regeneration.

## Docs sync

- `src/cli/AGENTS.md` + `skills/plexi-cli/SKILL.md` updated in the same PR (CLI skill-sync rule); `.agents/skills/babysitter/SKILL.md` got deprecation notes only — the full recipe sweep is task 0589.
- Regenerated: `pgap.schema.json`, `_protocol.py`, `cli.md`, `pgap.md`. All `just check-*` docs/schema gates green.

## Test evidence

- `cargo test --bin plexi` (env-stripped): **2056 passed; 0 failed; 3 ignored** with `headless_frame_fails_fast_when_the_guest_dies_at_import` skipped — that test is a pre-existing load flake that fails the full suite on clean alpha `a973ae47` too (evidence recorded in task 0580); it passes in isolation on this branch.
- New coverage: harness tests for all three verbs (settle/confirm, one-retry collapsed-paste heal, in-flight refusal, app-pane refusal, boot defer/timeout/pane-closed, alias typing, hidden-window servicing for each), plus CLI-side exit-code contract tests and wire-payload socket tests.
- Conclusion: binary install required — live TUI submit/boot behavior is the tester's round (`just pr-install <PR>`).


ianjamesburke:
**Fix round: boot-blindness for Codex resolved (8c69c28e).**

Root cause (verified by PTY probe against the real codex 0.145.0 binary with a hook-shim on PATH): Codex defers its `session_start` hook until the first user prompt submission — boot → 20s idle → zero hook invocations; the deferred idle report fires only alongside the first prompt. Claude Code fires SessionStart at boot, hence the asymmetry. The mid-run codex sightings were PreToolUse/PostToolUse/Stop, which Codex does emit.

Fix extends the ONE shared detector (no special case in the 0584 path, no prompt-signature regex): `tick_terminal_activity` synthesizes host-observed agent state when a known agent binary (kernel `p_comm`: codex/claude/pi) leads the PTY foreground group and no hook has reported — identity from the process name, idle/working from the PTY output settle. `Pane::agent()` merges hook reports (authoritative, permanent once any fires) over the synthesis, so `pane list`, `pane state`, and the boot predicate became correct together — including hand-spawned codex panes, which fixes the standalone `agent: null` bug too.

Headless repro of the tester's exact failure: `agent_spawn_completes_from_observed_codex_boot` (a real binary named `codex` in a real PTY shell, boot completes with zero hook reports), plus hidden-window, streaming-never-idle, and hook-overrides-observed harness tests.

Suite (env-stripped, known 0580 flake skipped + isolation pass 11.38s): `2063 passed; 0 failed; 3 ignored; 1 filtered; 104.82s`. All docs/schema gates green — no generated-artifact deltas (host-internal change only).

ianjamesburke:
## Fix round 2 — badd8256

**Root cause (identity, not timing):** `get_pid_name` used `libc::proc_name`, and macOS names a process exec'd through a symlink after the **resolved target**. The Homebrew codex cask is `codex -> codex-aarch64-apple-darwin`, so the detector saw `codex-aarch64-apple-darwin`, `known_agent_process` never matched, and the synthesis produced `None` every tick — exactly matching the log signature (fg activity recorded, no observed-agent transition). Verified live on pid of a running codex 0.145.0: `proc_name`/`p_comm`/`proc_pidpath` all report the payload name; only `KERN_PROCARGS2`'s saved exec path preserves `/opt/homebrew/bin/codex`.

**Why round 1's test lied:** the fake copied a binary *to the name* `codex`, guaranteeing the match the real symlinked install never produces — and a unit test asserted the payload-name non-match as correct.

**Fix:** `get_pid_name` now returns the basename of the as-invoked exec path (`KERN_PROCARGS2`), falling back to `proc_name`. No change to the settle logic, boot service, or 0584 path.

**Settle heuristic re-validated against the real binary:** real `codex -s danger-full-access -m gpt-5.6-terra` in a PTY — last output at **t=0.500s**, then **19.5s dead quiet**, max inter-burst gap 152ms. No idle spinner/repaint; the 1s settle window is structurally sound for this TUI.

**Tests that would have caught this:** `get_pid_name_reports_the_invoked_name_through_a_symlink` (unit) and `fake_idle_codex` now reproduces the real distribution shape (payload binary reached through a `codex` symlink) — both fail against round-1 code. Residual gap, stated honestly: the suite cannot depend on the real `/opt/homebrew/bin/codex` (external binary, CI); the risk left uncovered is a future codex version changing its idle output behavior. Today's behavior is pinned by the live measurements above.

**Finding 2 (idle-while-working) — not this PR, tracked as stint 0618:** post-hook display is byte-identical to alpha (`Pane::agent()` reads `t.agent` first; synthesis is disabled once any hook fires — the observed path is dead code after the first report). Headless baseline with the real codex + shimmed hook script: `UserPromptSubmit -> working` dispatched within ~1s of submission on second prompts and multiline prompts; could not reproduce the stall headless. Filed **stint 0618** for the intermittent hook under-reporting with repro rig details.

**Suite:** `2064 passed; 0 failed; 3 ignored; 1 filtered out; 89.06s` (env-stripped, memory-watchdog clean, known 0580 flake skipped). No generated artifacts touched (host-internal diff).

PR #2522, opened 2026-07-29
fix: stamp pane context env explicitly (stint 0607)

ianjamesburke:
## Summary

Implements stint 0607. No linked GitHub issue — stint is authoritative.

- Pass `PaneContextEnv` explicitly through root-pane, page, and terminal spawn seams.
- Resolve top-level context identity before its PTY is spawned, including path-backed contexts.
- Add a harness regression that inspects the environment handed to the PTY spawn.

## Done When

Pane environment is stamped from the pane's resolved context rather than `router.active()` or `active_window`; context creation and pane-spawn paths carry explicit identity. Existing panes are not re-stamped after moves.

**Test evidence (attempt 1):**
- cargo test: 2123 passed, 0 failed, 5 ignored — full bin suite; targeted `app::tests::context_tests::new_context_root_pty_env_uses_its_own_context_identity` also passed.
- Conclusion: install skippable — full headless harness coverage of the PTY-spawn environment contract.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1762,8 +1762,9 @@ impl PlexiApp {
         // workspace path returns earlier and is untouched. The welcome screen
         // still renders for any genuinely-empty window elsewhere.
         let seed_cwd = app.windows[0].path.clone();
+        let seed_context = app.pane_context_env_for_window(0);
         if app
-            .seed_window_root_pane(0, seed_cwd, None, false)
+            .seed_window_root_pane(0, &seed_context, seed_cwd, None, false)
             .is_some()
         {
             log::info!("first boot: seeded base root pane in context 1");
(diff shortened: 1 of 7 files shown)


ianjamesburke:
PARKED (unmerged) at operator wind-down for the beta v0.2.2 host restart — not a rejection.

Tester round 1 (live pr-2522 build, id-vs-name detail): **FAIL** — `new-context panes export generated "Context N" instead of their listed context name`. The context **id** stamping is not what failed here; the reported mismatch is on `PLEXI_CONTEXT_NAME`, where the spawn env carries the generated placeholder name while `pane list` shows the context's listed name.

Two things the next head must resolve before ordering any fix:
1. Rule whether the name discrepancy is **introduced by this PR or pre-existing on alpha** (baseline repro first). If the generated-vs-listed name divergence predates the diff, it is a follow-up stint, not a gate on #2522.
2. The tester found the workspace CLI does **not** expose `context new`; it validated via `context sub` instead. Confirm which creation paths are actually reachable from the CLI before writing the next pass criterion.

No fix round was opened. Resume via /babysitter resume.

PR #2531, opened 2026-07-31
Reclaim channel-scoped app state (stint 0648)

ianjamesburke:
## Summary

Implements stint 0648. No linked GitHub issue — stint is authoritative.

- Keeps `.plexi/app_states` as the deliberate channel-neutral state tier while leaving config channel-scoped.
- Reclaims the newest local or legacy global `.plexi-<channel>` orphan with no-clobber, verified migration; failures are surfaced instead of falling through to empty state.
- Makes state writes atomic and ensures context roots ignore `.plexi/app_states/` without overwriting user gitignore rules.

## Done When

- Existing state written under one channel is readable under another channel.
- Local and legacy global orphans are reclaimed idempotently without overwriting canonical state or deleting unverified source data.
- Missing state remains an empty first launch; unreadable, invalid, or non-object state returns a typed visible error.
- Assigning an existing context root ensures `.plexi/.gitignore` covers `app_states/`.

**Test evidence (attempt 1):**
- Focused migration tests: 10 passed, 0 failed — filter: `python_state_`
- Gitignore tests: 2 passed, 0 failed — filters: `app_state_gitignore_preserves_rules_and_is_idempotent`, `auto_init_existing_channel_workspace_ensures_neutral_app_state_ignore`
- cargo build: passed
- cargo test: 2134 passed, 0 failed, 5 ignored — full bin suite, env-stripped, RSS-watchdog guarded
- Fleet-load note: the first full run had one pre-existing 15-second guest-death timing failure; its quiet isolated rerun passed, then the final quiet full suite passed.
- Conclusion: binary install required — host/runtime filesystem migration must be validated by the separate tester against an installed PR channel.
- PR install: required via `just pr-install <PR>` by the tester; not run in Worker Mode.

## Review

Two Codex review rounds completed. The legacy global fallback gap and atomicity/race findings were fixed; no unresolved findings remain.


ianjamesburke:
## Fix round: multi-orphan reclaim (tester finding)

**Root cause:** the migration reasoned about one source, not the set — `adopt_python_state_orphan` popped the first entry of the sorted orphan list and dropped the rest, so with multiple channel orphans only the newest was adopted and older ones stayed in place looking live.

**Fix (16194ac2):** `reclaim_python_state_orphans` now reconciles the complete orphan set per tier on every load: canonical missing → newest adopted (verified copy, source removed); every remaining orphan is superseded and disposed of, including orphans that appear after a canonical file already exists — the category, not the instance.

**Built both** per standing directive, live-switchable via `[app_state] superseded_orphans`:
- `archive` (**default** — never destroys user bytes): rename to inert `<app_id>.json.superseded`, numbered on collision, hard-link-then-remove so an earlier archive is never clobbered.
- `remove`: delete only when byte-identical to canonical (proven copied); divergent orphans are archived even under `remove`.

Two invariants hold under both policies: never delete unproven-copied bytes; never touch a newer destination. Cleanup stint **0659** files deletion of the losing variant.

**Verification:** automated repro of the exact tester scenario (`python_state_reclaims_every_orphan_adopting_only_newest`: .plexi-beta older + .plexi-pr-999 newer → newest adopted, older archived, reopen idempotent), plus `python_state_remove_policy_deletes_only_proven_copies` and `python_state_archive_never_clobbers_prior_archive`. Full suite env-stripped: 2134 passed, 2 failed — both failures (`headless_frame_fails_fast_when_the_guest_dies_at_import`, `pane_send_submit_unconfirmed_reports_observed_input_line`) fail identically on the unmodified baseline aaa60348 under tonight's machine load, so they are environmental, not from this diff. Config docs regenerated.

PR #2537, opened 2026-07-31
DO NOT MERGE: alpha test-suite CI baseline, zero skips (stint 0649)

ianjamesburke:
Measures which tests fail on clean alpha in the CI environment with no skips. Workflow file is the only change (the measuring instrument). Will be closed and deleted after the run. Never merge.

### DIFF
--- a/.github/workflows/rust-host.yml
+++ b/.github/workflows/rust-host.yml
@@ -1,0 +1,44 @@
+name: Rust host
+
+on:
+  pull_request:
+    branches: [alpha]
+
+env:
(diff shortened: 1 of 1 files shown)


ianjamesburke:
Baseline captured: clean alpha in CI with zero skips fails exactly one test, host::wasm_python::tests::headless_frame_fails_fast_when_the_guest_dies_at_import (timeout backstop instead of process-exit death path). 2122 passed, 0 filtered. Both pane_send_submit tests pass in CI; the two send_to_pane/pgap skip names match no test in the tree.

PR #2564, opened 2026-08-02
fix: polish sidebar context rows (stint 0700)

ianjamesburke:
## Summary

Implements stint 0700. No linked GitHub issue; stint is authoritative.

- Aligns the context number with its headline, pins notification counts to the trailing edge, and raises only the headline token to 14 px.
- Ships the close affordance as `X: end`: the X follows the notification count on the active row. The inline candidate and its toggle are deleted.
- Restores the default sidebar width to 220 points. The row finds headline room internally by tightening the number gutter, one-digit count reserve, X hit target, and title-to-pips gap. `PLEXI Workspace` fits at 220; the deliberately long fixture ellipsizes.
- Sidebar text size is not configurable today. The existing terminal font size and pane-title font size do not control sidebar rows; this stint deliberately adds no config key.

## Done When

- Context number aligns with the headline.
- Notification count is trailing-aligned without overlap.
- Headline is larger while subtitle/body text remains unchanged.
- `X: end` is the sole shipped close affordance.

**Test evidence:**

- `env -u PLEXI_SOCKET cargo test --bin plexi screenshot_sidebar_x_end_at_default_width -- --nocapture`: passed.
- Re-rendered current-state evidence: `.stint/evidence/0700-sidebar-x-end.png`. The fixture includes `PLEXI Workspace` and `Narrator Video Production Workspace`.

### DIFF
--- a/src/app/render.rs
+++ b/src/app/render.rs
@@ -3,6 +3,7 @@
 use super::{ClickFlash, PlexiApp};
 use crate::app_protocol::AgentState;
 use crate::spatial::tiling::{PaneId, PlexiBehavior};
+use crate::ui::style;
 use egui::{Color32, CornerRadius, Stroke, StrokeKind, Vec2};
 use egui_tiles::{Tile, TileId};
 use std::collections::HashMap;
@@ -187,7 +188,7 @@ impl PlexiApp {
                 }
             });
             egui::Panel::left("sidebar")
-                .default_size(220.0)
+                .default_size(style::SIDEBAR_DEFAULT_W)
                 .size_range(140.0..=400.0)
                 .resizable(true)
                 .show_separator_line(false)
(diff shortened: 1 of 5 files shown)


ianjamesburke:
Re-check note (babysitter head): the targeted re-check found that at the host's 400-point minimum window width, the new 240-point sidebar default leaves 160 points for the central content area. Ruled **not blocking for this PR**: the condition is pre-existing rather than introduced here — the previous 220-point default left 180 points at the same window width, equally unusable. The actual defect is that the sidebar has no responsive clamp at any width, filed as stint 0713.

The re-check also confirmed: no stale 220-point assumptions remain anywhere in the tree, existing persisted panel widths are preserved (240 applies only to fresh panel state), install provenance proven at 9698034f, context selection works, and the focused close-path test passes.

PR #2569, opened 2026-08-02
fix(ui/sidebar): single-context sidebar inflates to max width with a dead gap strip and un-resizable panel (stint 0715)

ianjamesburke:
Closes stint **0715** — *fix(ui/sidebar): single-context sidebar inflates to max width with a dead gap strip and un-resizable panel*. No linked GitHub issue; stint is authoritative.

## Root cause (measured, not assumed)

The stint left the root cause open. A frame-over-frame probe pinned it exactly:

```
[row] outer_avail=198 text_max=198 title_w=29.6 pip=0 badge=0 close=0 min_rect_w=206 spacing=8
[probe] frame=0 panel_w=228   frame=1 236   frame=2 244 ... +8pt every frame
```

`ContextItem`'s headline lays out `label → pad-to-budget → pips → trailing block` and budgets each of their **widths**, but never egui's implicit `item_spacing.x` between them. The row allocated `available_width + 8`.

That alone would just clip. What made it compound: egui stores a panel's **content** rect as the panel's size (`PanelState { rect }.store` on the frame's response rect), and `PanelSizer::get_size_from_state_or_default` reads that back as the panel's width next frame. So the overflow fed itself — a monotonic ramp to the `size_range` 400-point cap, which also explains the un-resizable handle (a drag is overwritten the very next frame) and the dead strip (the panel's `clip_rect` stays at `panel_rect` while `cursor.min.x` for the central panel is advanced to the wider content rect, so the difference is painted by neither).

**Why a second context "fixed" it** (the stint's other open question): it does not. Two contexts enable the close zone, which shifts 22pt out of the text budget but leaves the same single unbudgeted 8pt gap — it still overflows. What changes is that the ramp starts from an already-clamped state in the reporter's session; the geometry test in this PR shows a 2+-context row overflowing pre-fix too (`panel_w=160 row_w=168` with `action_enabled=true`). So this was never a single-context-only defect; the first-run state is simply where it is most visible, because a lone context suppresses the close zone and gives the name its widest budget.

## Fix

Turn implicit item spacing off on the headline row and budget every gap explicitly:

- the pip/text gap becomes a named `PIP_TEXT_GAP` and an explicit `add_space`, matching what `pip_strip_w` already reserved;
- the trailing right-to-left block keeps normal spacing internally, and that gap is now part of `trailing_w`.

Row width now equals the panel's available width exactly, for every combination of close zone, badge, and pip count. Visual design from stint 0700 is unchanged.

Out of scope, as the stint specifies: the narrow-window responsive clamp / content-area floor (stint 0713).

## Test evidence

- **`ui::sidebar_row::tests::context_row_never_exceeds_available_width`** — sweeps panel width × close zone × badge count × pip count and asserts the row never claims more than it was given.
- **`testing::harness_tests::sidebar_panel_width_tests::single_context_sidebar_settles_at_default_width`** — first-run state settles at `SIDEBAR_DEFAULT_W` (220) and does not grow across 20 further frames.
- **Falsified**: both go red with the fix reverted — `row overflowed: panel_w=160 row_w=168`, and `left: 236.0 right: 220.0`.
- **`ui_tests::tests::screenshot_single_context_sidebar`** — visual review surface at ppp 2.0. Rendered and inspected: sidebar at 220pt, row card fully inside the panel with its 4pt inset, no dead strip between sidebar and central pane. PNG deleted after review per TESTING.md.
- **Full suite**: `cargo test --bin plexi` — 2243 passed, 3 failed. All 3 (`cargo_lease_serializes_concurrent_invocations`, `cargo_lease_sigkilled_wrapped_command_reports_137`, `wasm_python::headless_frame_fails_fast_when_the_guest_dies_at_import`) are self-interference from running the suite *inside* `scripts/cargo-with-lease.sh` plus load-sensitive timeouts; re-run outside the wrapper all 7 in that filter pass. Unrelated to this diff.
- **Codex review** (`--base alpha`): clean, no findings.
- **Conclusion**: install skippable — the geometry invariant, the frame-over-frame settling behavior, and the rendered pixels are all covered headlessly. A `just pr-install` pass is still worth it if a human wants to confirm the resize handle feels right by hand, since drag feel is the one thing the harness does not reproduce.

### DIFF
--- a/src/testing/harness_tests.rs
+++ b/src/testing/harness_tests.rs
@@ -6932,3 +6932,44 @@ fn logic_drains_run_without_panic_visible_and_hidden() {
         "hidden frame pass must still run every logic drain and end at logic_complete"
     );
 }
+
+/// Stint 0715: the sidebar's context row used to allocate `available_width`
+/// plus one unbudgeted `item_spacing.x`. egui stores a panel's *content* rect
+/// as that panel's size, so the overflow was read back as the panel width on
+/// the next frame and re-overflowed — a monotonic ramp (+8pt/frame here) that
+/// pinned the sidebar at its `size_range` maximum and undid any user resize.
+/// The first-run state, exactly one context, is the one that showed it.
+#[cfg(test)]
+mod sidebar_panel_width_tests {
+    use super::*;
+
+    fn sidebar_width(h: &HostHarness) -> f32 {
+        egui::containers::panel::PanelState::load(&h.ctx, egui::Id::new("sidebar"))
+            .expect("sidebar panel state must exist after a frame")
+            .rect
+            .width()
+    }
+
+    #[test]
+    fn single_context_sidebar_settles_at_default_width() {
+        let mut h = HostHarness::new();
+        h.app.sidebar_visible = true;
+        assert_eq!(h.app.router.len(), 1, "setup: first-run is one context");
+
+        h.run_frames(2);
+        let settled = sidebar_width(&h);
+        assert_eq!(
+            settled,
+            crate::ui::style::SIDEBAR_DEFAULT_W,
+            "a single-context sidebar must settle at SIDEBAR_DEFAULT_W"
+        );
+
+        h.run_frames(20);
+        assert_eq!(
+            sidebar_width(&h),
+            settled,
+            "the sidebar must not grow frame over frame — a widening panel means \
+             the context row is again allocating more than the panel's width"
+        );
+    }
+}
--- a/src/ui/AGENTS.md
+++ b/src/ui/AGENTS.md
@@ -9,7 +9,7 @@ Host overlays should use these instead of raw egui layout wherever a primitive e
 - `overlay::ModalShell` for modal frame, scrim, title, body, scroll body, and dismissal.
 - `typography` for modal titles, section labels, body text, captions, and muted text.
 - `list::ListRow`, `row`, `text_field::TextField`, `button`, `hints::HintBar`, and `surface` for repeated chrome.
-- `sidebar_row::SidebarRow` for context-list rows — index gutter, badge slot, close action, and per-window pane-pip capsules. It resolves every slot rect in one `measure` pass; never hand-place a piece of a sidebar row.
+- `sidebar_row::SidebarRow` for context-list rows. Two tiers, the same convention `ListRow` uses for a secondary line: identity on top (index gutter, name, close action), pane state below (per-window pip capsules, overflow count, notification badge, root path). It resolves every slot rect in one `measure` pass; never hand-place a piece of a sidebar row.
 
 ## Design Tokens (`src/style.rs`)
 
--- a/src/ui/AGENTS.md
+++ b/src/ui/AGENTS.md
@@ -9,6 +9,7 @@ Host overlays should use these instead of raw egui layout wherever a primitive e
 - `overlay::ModalShell` for modal frame, scrim, title, body, scroll body, and dismissal.
 - `typography` for modal titles, section labels, body text, captions, and muted text.
 - `list::ListRow`, `row`, `text_field::TextField`, `button`, `hints::HintBar`, and `surface` for repeated chrome.
+- `sidebar_row::SidebarRow` for context-list rows — index gutter, badge slot, close action, and per-window pane-pip capsules. It resolves every slot rect in one `measure` pass; never hand-place a piece of a sidebar row.
 
 ## Design Tokens (`src/style.rs`)
 
--- a/src/ui/sidebar.rs
+++ b/src/ui/sidebar.rs
@@ -1,7 +1,7 @@
 use crate::host::context::WindowMenuAction;
 use crate::ui::button;
 use crate::ui::list::ListDropdownHeader;
-use crate::ui::sidebar_row::{ContextItem, PaneDots, SidebarAction};
+use crate::ui::sidebar_row::{PaneDots, SidebarAction, SidebarRow};
 use crate::workspace::router::ContextMove;
 use egui::{Align, CornerRadius, Layout, Rect, RichText, Stroke, Vec2};
 use egui_tiles::Tile;
@@ -227,7 +227,7 @@ impl PlexiApp {
             let ctx_id = self.router.get(i).context_id;
             let pane_dots = self.sidebar_pane_dots(ctx_id, is_active);
 
-            // --- Renaming: special-cased before ContextItem path ---
+            // --- Renaming: special-cased before the SidebarRow path ---
             if is_renaming {
                 let fill = if is_active {
                     self.colors.bg_active
@@ -295,7 +295,7 @@ impl PlexiApp {
                 continue;
             }
 
-            // --- Normal row via ContextItem ---
+            // --- Normal row via SidebarRow ---
             let ctx_name = self.router.get(i).name.clone();
             let badge_count = if is_active {
                 self.visible_notification_count()
@@ -304,7 +304,7 @@ impl PlexiApp {
             };
             let subtitle = Some(self.router.get(i).root.display().to_string());
 
-            let (action, response) = ContextItem {
+            let (action, response) = SidebarRow {
                 is_active,
                 is_dragging,
                 any_dragging,
@@ -316,7 +316,7 @@ impl PlexiApp {
                 pane_dots,
                 draggable: true,
             }
-            .draw(ui, egui::Id::new(("ctx", i)), &self.colors);
+            .show(ui, egui::Id::new(("ctx", i)), &self.colors);
 
             active_rects.push(response.rect);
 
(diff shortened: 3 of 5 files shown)


ianjamesburke:
## Stint 0717 — context row rebuilt on a measured primitive

Pushed `821edfea` onto this branch.

### Shape
Neither wholesale option in the spec fit, so I took the second one and named it for symmetry with the existing primitive: **`SidebarRow` in `src/ui/sidebar_row.rs`**, built from `list.rs`'s shared internals (`selection_shapes`/`paint_selection`, `selection_inset`, `elide_to_width`, `paint_text_centered`, `snap::*`). Grafting an index gutter, a badge slot, drag sensing, and windowed pip capsules onto `ListRow` would have added four single-caller features to a primitive used by every list surface in the host, and `ListRow` is fixed-height by design while the context row is not.

`RowGeometry::measure` resolves the entire row in one pass into explicit slot rects in row-local space; `translated` moves them onto the allocated rect. Painting and hit testing both read those rects — nothing is reconstructed afterwards.

### What went away
- `name_row_h_capture` / `close_rect_capture` and the post-scope index painting.
- The hand-rolled headline budgeting (`close_w`/`badge_w`/`pip_strip_w` with `item_spacing` zeroed and gaps re-added by hand).
- The duplicated pip-width math — `PipLayout::measure` is now the only place it exists, used both for the title budget and for painting.
- The `max()` fudge clamp on the pin triangle.

### What changed behaviorally
- **Trailing slots** are laid out right to left at fixed widths (close, then badge); they cannot overlap or swap at any panel width. The close glyph paints into the same rect the pointer is tested against.
- **Index gutter** is centered on the title line, not the row, so the number shares the name's optical center.
- **Pip strip** shrinks its visible dot count to fit its budget and folds the rest into `+N`, instead of running under the name at narrow widths. The name keeps a floor (`TITLE_MIN_W`).
- **Pin**: base flush on the capsule's inner top edge (`min.y + stroke.width`), tip one point short of the dots, both snapped to the same pixel grid. Two `const _: () = assert!` statements freeze that contract at compile time.
- **Accessibility**: the row paints its own text now, so it reports the full un-elided context name via `Response::widget_info`. Without this, `host_has_label` and scene `assert_label` stop seeing context names — two existing tests caught it.
- Content with no trailing slot in front of it keeps the row's right margin, so a long root path no longer touches the selection card's outline.

### Files
- `src/ui/sidebar_row.rs` — rewritten
- `src/ui/sidebar.rs` — call sites (`ContextItem` → `SidebarRow`, `.draw` → `.show`)
- `src/ui_tests.rs` — new screenshot surface + seeding helpers
- `src/ui/AGENTS.md` — `SidebarRow` listed under primitives

### Tests
`cargo build` clean. `cargo test --bin plexi`: **2252 passed, 0 failed, 5 ignored** (no flakes on this run). `cargo clippy --bin plexi --all-targets` reports nothing in the touched files.

Both guarded invariants green:
- `ui::sidebar_row::tests::context_row_never_exceeds_available_width`
- `testing::harness_tests::sidebar_panel_width_tests::single_context_sidebar_settles_at_default_width`

New geometry tests: `trailing_slots_never_overlap_at_any_width` (5 panel widths × close × badge × pane-count), `index_gutter_is_centered_on_the_title_line`, `pip_strip_width_is_measured_once`, `each_spatial_window_gets_its_own_capsule`, `pips_focused_pane_past_dot_cap_does_not_panic`.

### Visual review
`ui_tests::tests::screenshot_sidebar_row_states` (new, ppp 2.0) renders three seeded contexts: an active row with two window capsules (3 + 2 dots, agent activity colors, accent stroke on the active window), a long elided name with 5 dots + `+6` overflow and a badge, and a hovered row showing its close glyph. I read the PNGs at 8× magnification and confirmed: the index number sits on the name's optical center on every row; close / badge / pips are three separated slots with clear air between them; the capsule's top stroke is unbroken except where the pin meets it, with the pin's base flush on the edge and its tip pointing at the focused dot. The single-context shot (`screenshot_single_context_sidebar`) still fits inside the 220pt panel with no clipping. PNGs deleted after review.

Note for whoever tests this: `just pr-install` not run, and `stint done` deliberately not called.

ianjamesburke:
## Two-tier context row (iteration on Ian's 5:39pm review)

Pushed `b9d14d70`.

### Layout
**Identity tier** — index gutter, context name, close X pinned to the top-right corner at `TEXT_TITLE` (up from `TEXT_SIDEBAR_TITLE`) in a 26pt hit slot. Nothing else on this line. The name's budget is now one subtraction from the close slot and doesn't move when pane state changes.

**Pane-state tier** — pip capsules lead at the content lane, then `+N` overflow, then the notification badge, then the root path. Fixed reservations left to right; the path is the only elastic element, so no pair can collide.

This is `ListRow`'s convention (secondary line carries the metadata), applied through the same shared internals rather than duplicated. I did not extend list.rs itself: `ListRow` right-aligns its metadata lane, while Ian's spec puts the pips left-aligned after the gutter, so the shared thing is the tiering, not the lane placement — forcing both through one code path would have meant a mode flag on `ListRow` for one caller.

### The path's fate
Kept, trailing the pane-state tier, elided to whatever the pips and badge leave. **Dropped entirely below 48pt of remaining lane** rather than shown as `/pro…`, which is noise. That's a content rule inside one already-measured lane, not a second row shape — the geometry is identical either way.

### Badge vs collapsed pips (the bug)
Two changes, either of which alone would have fixed it:
- The `+N` overflow now paints in a **dot's own color** via `activity::pip_color` — the same gray as the dots it stands in for, or the focused-dot color when the collapsed range holds the focused pane. It no longer borrows the accent the badge owns. (The old code deliberately turned the overflow accent-colored when the focused pane was past the cap, which is what made it a badge lookalike.)
- The badge is now a **filled accent pill** with `text_on(accent)` for contrast — taller than a pip, different shape, different color family, and separated by `SPACE_SM` so it never reads as glued to the strip.

### Other
- The close slot overhangs into the row's vertical padding instead of stretching the identity tier, so the bigger glyph costs zero row height.
- A row with no pane state at all is one tier tall (no stray tier gap).
- Content right margin moved to `SPACE_MD`, so text has visible air inside the selection card's outline instead of nearly touching it.

### Tests
`cargo test --bin plexi`: **2252 passed, 1 failed → the known flaky `headless_frame_fails_fast_when_the_guest_dies_at_import`, which passes in isolation** (rerun: 1 passed). 5 ignored. Clippy and rustfmt clean on all touched files.

0715 invariants green: `context_row_never_exceeds_available_width`, `single_context_sidebar_settles_at_default_width`.

Replaced `trailing_slots_never_overlap_at_any_width` with `tiers_never_overlap_at_any_width` — walks the pane-state tier left to right across 5 panel widths × close × badge × pane-count, asserting each element clears its predecessor, stays inside the row, and sits below the identity tier, plus that the close action stays on the identity tier and never runs into the name. Added `badge_is_visually_distinct_from_collapsed_pips`, which freezes both the gap and the pill-taller-than-pip relationship against a row whose pips have actually collapsed.

### Visual review
`screenshot_sidebar_row_states` re-seeded and read at 8×: active row shows two window capsules with activity colors and the return pin on line 2 followed by the accent badge and the path; the overflow row shows 8 dots + a gray `+3` + a blue pill `2` that are now obviously three different things; the hovered row shows the larger X at its top-right. The single-context first-run shot still fits the 220pt panel with no clipping. PNGs deleted after review.

ianjamesburke:
## Spacing pass (iteration on Ian's 6:26pm review)

Pushed `a510816e`. All four asks, tokens where they fit:

| | before | after |
|---|---|---|
| Row vertical padding | `7.0` | `style::SPACE_MD` (12) |
| Gap between the tiers | `3.0` | `style::SPACE_SM` (8) |
| Capsule interior padding (x / y) | `5` / `4` | `7` / `6` |
| Capsule → `+N` gap | `SPACE_XS` (4) | `SPACE_SM` (8) |

Row height goes from ~52pt to ~71pt, which is the trade Ian OK'd. The capsule's interior padding stays as named local constants rather than being forced onto the spacing scale — `SPACE_SM` on both axes would have made a 24pt-tall capsule next to a 19pt title, which is bulkier than the dots need.

One structural change came out of it: **`PIN_H` is now a fixed `3.0` instead of `WINDOW_GROUP_PAD_Y - 1.0`.** Tying it to the padding meant the return-target marker grew every time the capsule got roomier — a 5pt caret at the new padding. The tip's clearance over the dot row was never really a function of the padding anyway; the `const _: () = assert!(PIN_H < WINDOW_GROUP_PAD_Y)` is what enforces it, and it still does.

The index gutter is untouched, as asked. It derives from the title's optical center, so it followed the new padding without any adjustment.

### Tests
`cargo test --bin plexi`: 2252 passed, 1 failed — `pane_send_submit_unconfirmed_reports_observed_input_line`, which passes in isolation and on a full module rerun (9/9). It's a real-PTY timing test of the class TESTING.md flags for load sensitivity, and it shares no code path with this diff (the commit touches only `src/ui/sidebar_row.rs`). Not one of the three known flakies, so flagging it explicitly rather than waving it through.

0715 invariants green. Clippy and rustfmt clean.

### Visual review
Re-rendered both screenshot tests and read them at 8×. The tiers now read as two separate statements instead of a wrapped line; the dots sit in air inside their capsules with a visible gap to the stroke; the pin is a small caret with clear space above the focused dot; `+3` has real separation from the capsule and is still obviously a different kind of thing from the blue badge pill next to it. The single-context first-run row still fits the 220pt panel with no clipping. PNGs deleted.

